PR #4, opened 2016-04-28
Goa 1993 index annotation data

eddturner:
Could someone review this please? It's for annotation indexing. Thanks!


indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationConfig.java
```diff
@@ -0,0 +1,192 @@
...
+    private JobExecutionListener logJobListener() {
+        return new LogJobListener();
+    }
+
+    private StepListener logStepListener() {
+        return new LogStepListener();
+    }
+
+    private StepListener skipLogListener() {
+        return new SkipLoggerListener();
```
TonyWardell: SkipLoggerListener has hardcoded references to GeneProductDocument.

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * Created 19/04/16
+ * @author Edd
+ */
+public class AnnotationDocumentConverter implements ItemProcessor<Annotation, AnnotationDocument> {
+    static final int DEFAULT_TAXON = 0;
+
+    @Override public AnnotationDocument process(Annotation annotation) throws Exception {
+        if (annotation == null) {
+            throw new DocumentReaderException("Annotation object is null");
```
TonyWardell: Is DocumentReaderException correct? It doesn't feel right for going into Solr, rather the reverse...

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        Map<String, String> propertiesMap = convertLinePropertiesToMap(annotation.annotationProperties, PIPE, EQUALS);
```
TonyWardell: There is no unit test on this class where annotationProperties are empty or null. 
Obviously this is handled by the statically called method - but are there any side effects to this class if this was the case?

indexing/src/main/java/uk/ac/ebi/quickgo/index/common/datafile/GOADataFileParsingHelper.java
```diff
@@ -0,0 +1,67 @@
...
+                            String[] splitProp = splitValue(unformattedProp, intraValueDelimiter);
+                            propMap.put(splitProp[0], splitProp.length == 2 ? splitProp[1] : "");
+                        });
+            }
+        }
+
+        return propMap;
+    }
+
+    public static String[] splitValue(String value, String delimiter) {
```
TonyWardell: How do you feel about replacing this code with the following?

    public static String[] splitValue(String value, String delimiter) {
        assert delimiter != null : "Delimiter cannot be null";

        return Optional.ofNullable(value)
                .filter(v -> !v.isEmpty())
                .map(v -> value.split(delimiter))
                .orElse(new String[0]);

    }

Lamda infected!? :)

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
```diff
@@ -0,0 +1,89 @@
...
+
+        AnnotationDocument doc = new AnnotationDocument();
+
+        doc.geneProductId = constructGeneProductId(annotation);
```
TonyWardell: Why concatenate the db and gene product id at this stage. Wouldn't it be better to keep them separate?

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
```diff
@@ -0,0 +1,89 @@
...
+        doc.qualifier = annotation.qualifier;
+        doc.goId = annotation.goId;
+        doc.reference = annotation.dbReferences;
+        doc.assignedBy = annotation.assignedBy;
+        doc.ecoId = annotation.ecoId;
+        doc.extensions = constructExtensions(annotation);
+
+        doc.withFrom = constructWithFrom(annotation);
```
TonyWardell: These two methods are the same - pass in the value you are checking and you could have just one method..

    private List<String> constructExtensions(Annotation annotation) {
        return annotation.annotationExtension == null ? null :
                Arrays.asList(annotation.annotationExtension.split(PIPE));
    }

    private List<String> constructWithFrom(Annotation annotation) {
        return annotation.with == null ? null : Arrays.asList(annotation.with.split(PIPE));
    }

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
```diff
@@ -0,0 +1,89 @@
...
+        doc.interactingTaxonId = extractInteractingTaxonId(annotation);
```
TonyWardell: Looks the same as GeneProductParsingHelper.extractTaxonIdFromValue.

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationValidator.java
```diff
@@ -0,0 +1,127 @@
...
+        if (!Strings.isNullOrEmpty(annotation.with) && !WITH_REGEX.matcher(annotation.with).matches()) {
+            handlePatternMismatchError(
+                    "With",
+                    annotation.with,
+                    WITH_REGEX.pattern(),
+                    annotation);
+        }
+    }
+
+    private void handlePatternMismatchError(String fieldName, Object fieldValue, String pattern, Object rawObject) {
```
TonyWardell: This method could be put into the common module. Add it to ValidationHelper?

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationParsingHelper.java
```diff
@@ -0,0 +1,49 @@
...
+            "(" + PIPE_SEPARATED_CSVs_FORMAT + ")|(With:Not_Supplied)",
+            DB_COLON_REF_FORMAT, DB_COLON_REF_FORMAT, DB_COLON_REF_FORMAT, DB_COLON_REF_FORMAT));
+    static final Pattern QUALIFIER_REGEX = Pattern.compile(QUALIFIERS_FORMAT);
+    static final Pattern ANNOTATION_EXTENSION_REGEX = Pattern.compile(String.format(
+            PIPE_SEPARATED_CSVs_FORMAT,
+            WORD_LBRACE_WORD_RBRACE_FORMAT, WORD_LBRACE_WORD_RBRACE_FORMAT,
+            WORD_LBRACE_WORD_RBRACE_FORMAT, WORD_LBRACE_WORD_RBRACE_FORMAT));
+    static final Pattern ANNOTATION_PROPERTIES_REGEX = Pattern.compile(String.format(
+            PIPE_SEPARATED_CSVs_FORMAT,
+            KEY_EQUALS_VALUE_FORMAT, KEY_EQUALS_VALUE_FORMAT, KEY_EQUALS_VALUE_FORMAT, KEY_EQUALS_VALUE_FORMAT));
```
TonyWardell: So there is always 4 values in properties?

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationValidator.java
```diff
@@ -0,0 +1,127 @@
...
+                        ANNOTATION_PROPERTIES_REGEX.pattern(),
+                        annotation);
+            }
+            else {
+                checkMandatoryPropertiesFieldsExist(annotation);
+            }
+        }
+    }
+
+    private void checkMandatoryPropertiesFieldsExist(Annotation annotation) {
```
TonyWardell: Shouldn't symbol always exist - its an always populated column, so we should check that in this method.

indexing/src/main/java/uk/ac/ebi/quickgo/index/common/listener/ItemRateWriterListener.java
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Computes the rate of items written per second, from instance creation time
+     * to a specified time-point, {@code now}.
+     *
+     * @param duration the duration for which the rate should be computed
+     * @param writeCount the number of items written
+     * @return a floating point number representing the rate of writing
+     */
+    float getItemsPerSecond(Duration duration, AtomicInteger writeCount) {
```
TonyWardell: Doesn't appear to be used.

indexing/src/main/java/uk/ac/ebi/quickgo/index/geneproduct/GeneProductParsingHelper.java
```diff
@@ -27,50 +24,6 @@
...
-        if (value != null && value.trim().length() > 0) {
-            splitValues = value.split(delimiter);
-        } else {
-            splitValues = new String[0];
-        }
-
-        return splitValues;
-    }
-
     static int extractTaxonIdFromValue(String value) {
```
TonyWardell: So this method is more to do with taxons than it is gene products.

indexing/src/test/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverterTest.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    // interacting taxon
+    @Test
+    public void convertsEmptyInteractingTaxonToDefaultTaxon() throws Exception {
+        annotation.interactingTaxonId = null;
+
+        AnnotationDocument doc = converter.process(annotation);
+
+        assertThat(doc.interactingTaxonId, is(DEFAULT_TAXON));
```
TonyWardell: I think the interacting taxon should stay null if its not in the import file - as there is no interaction. Please check with Tony S.

indexing/src/test/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverterTest.java
```diff
@@ -0,0 +1,245 @@
...
+        assertThat(doc.interactingTaxonId, is(12345));
+    }
+
+    @Test
+    public void convertsInvalidNonEmptyInteractingTaxon() throws Exception {
+        annotation.interactingTaxonId = "taxon:12345d";
+
+        AnnotationDocument doc = converter.process(annotation);
+
+        assertThat(doc.interactingTaxonId, is(DEFAULT_TAXON));
```
TonyWardell: Same for this one.

indexing/src/test/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverterTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+        AnnotationDocument doc = converter.process(annotation);
+
+        assertThat(doc.taxonId, is(DEFAULT_TAXON));
+    }
+
+    // annotation properties: db object type
+    @Test
+    public void convertsNullDbObjectTypeAnnotationPropertiesToNullValue() throws Exception {
+        annotation.annotationProperties = null;
```
TonyWardell: DbObjectType should never be null.

indexing/src/test/java/uk/ac/ebi/quickgo/index/annotation/AnnotationValidatorTest.java
```diff
@@ -0,0 +1,398 @@
...
+
+    @Before
+    public void setUp() {
+        validator = new AnnotationValidator();
+        annotation = createValidAnnotation();
+    }
+
+    // check required fields -------------------------------------------------
+    @Test(expected = DocumentReaderException.class)
+    public void nullGeneProductThrowsException() throws Exception {
```
TonyWardell: Cut and paste ;)

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
eddturner: empty/null annotationProperties values are tested in GOADataFileParsingHelperTest (as you mention), but the validation of the values takes place in AnnotationValidator (and has accompanying unit tests).

indexing/src/main/java/uk/ac/ebi/quickgo/index/common/datafile/GOADataFileParsingHelper.java
eddturner: Sure, though with 1 minor change ...

```java
return Optional.ofNullable(value)
            .filter(v -> !v.isEmpty())
            .map(v -> v.split(delimiter)) // transform v (not value), to protect against problems with future refactoring
            .orElse(new String[0]);
```

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
eddturner: * Current requirements suggest this isn't necessary
  * Also, we could use a solr copy-field with a certain analyzer to do this for us, if necessary
* So far, IDs specified from users have come as <RESOURCE>:<ID> (e.g., when they type in an end-point)

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationParsingHelper.java
eddturner: Properties can contain any number of values. However, annotation requires these 4 required ones.

indexing/src/test/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverterTest.java
eddturner: `AnnotationDocument`'s field, `interactingTaxonId`, is an `int` so can't be null -- instead the value of `0` indicates it wasn't specified as having an interacting taxonomy ID.

indexing/src/test/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverterTest.java
eddturner: Yes. Though I thought it was a good idea to leave this test in place, in case in future we decide to omit the validation (`AnnotationValidator`) step. If we do so, then this just makes sure the system doesn't completely bail out and does something sensible in the circumstances.

indexing/src/main/java/uk/ac/ebi/quickgo/index/geneproduct/GeneProductParsingHelper.java
eddturner: I didn't write this -- but from the method name, GP data file contains extractable taxon info.

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
eddturner: `annotation` comes from reading the source files, and if `null` indicates something went wrong during reading.

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
eddturner: done

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationValidator.java
eddturner: no longer part of annotation documents

eddturner:
Hi Tony, I've commented / done each of your suggestions. Feedback would be great whenever is good. Cheers.

indexing/src/test/java/uk/ac/ebi/quickgo/index/annotation/AnnotationValidatorTest.java
eddturner: done

PR #6, opened 2016-05-16
Goa 1846 gene product load tests

eddturner:
* Converted existing ontology testing that needed running in stand-alone jmeter, to one using the jmeter-maven-plugins, so the performance tests can be run independently via maven only (therefore can be run headlessly, e.g., on jenkins, etc.).
* Created geneproduct tests focussing on ID single/batch retrieval; simple search queries; longer running search queries

* Caveats:
  * The tests written exercise desired behaviour. Not all the tests pass due to currently missing functionality, e.g., ID validation is not yet present -- therefore, ID searches for invalid IDs causes backend to work hard to try to find an ID that doesn't exist; and a 200 status is returned (with no results). When ID validation has been put in place, such tests should run quicker, return non 200 codes, and pass.


rantunesboreas:
Here are some issues I came across. I need to write them here because the files are too big to be shown in git review tool.

ids-10-per-line: last line, has only 1 accession.

geneproduct-rest/src/test/jmeter/geneproduct-rest-jmeter-tests.jmx
```diff
@@ -0,0 +1,1365 @@
...
+            </objProp>
+            <stringProp name="filename"></stringProp>
+          </ResultCollector>
+          <hashTree/>
+        </hashTree>
+        <HTTPSamplerProxy guiclass="HttpTestSampleGui" testclass="HTTPSamplerProxy" testname="Search on synonym" enabled="true">
+          <elementProp name="HTTPsampler.Arguments" elementType="Arguments" guiclass="HTTPArgumentsPanel" testclass="Arguments" testname="User Defined Variables" enabled="true">
+            <collectionProp name="Arguments.arguments">
+              <elementProp name="query" elementType="HTTPArgument">
+                <boolProp name="HTTPArgument.always_encode">false</boolProp>
```
rantunesboreas: I would suggest not using the synonym filed explicitly, and to use the /search request handler, setup to run queries on various fields, of which synonym is one of them.

geneproduct-rest/src/test/jmeter/geneproduct-rest-jmeter-tests.jmx
```diff
@@ -0,0 +1,1365 @@
...
+            <stringProp name="filename"></stringProp>
+          </ResultCollector>
+          <hashTree/>
+        </hashTree>
+        <HTTPSamplerProxy guiclass="HttpTestSampleGui" testclass="HTTPSamplerProxy" testname="Search by symbol" enabled="true">
+          <elementProp name="HTTPsampler.Arguments" elementType="Arguments" guiclass="HTTPArgumentsPanel" testclass="Arguments" testname="User Defined Variables" enabled="true">
+            <collectionProp name="Arguments.arguments">
+              <elementProp name="query" elementType="HTTPArgument">
+                <boolProp name="HTTPArgument.always_encode">false</boolProp>
+                <stringProp name="Argument.value">symbol:${SYMBOL}</stringProp>
```
rantunesboreas: Same comment here, as the one for synonym.

geneproduct-rest/src/test/jmeter/geneproduct-rest-jmeter-tests.jmx
```diff
@@ -0,0 +1,1365 @@
...
+            <intProp name="Assertion.test_type">5</intProp>
+          </ResponseAssertion>
+          <hashTree/>
+        </hashTree>
+      </hashTree>
+      <ThreadGroup guiclass="ThreadGroupGui" testclass="ThreadGroup" testname="Simple search requests" enabled="true">
+        <stringProp name="ThreadGroup.on_sample_error">continue</stringProp>
+        <elementProp name="ThreadGroup.main_controller" elementType="LoopController" guiclass="LoopControlPanel" testclass="LoopController" testname="Loop Controller" enabled="true">
+          <boolProp name="LoopController.continue_forever">false</boolProp>
+          <stringProp name="LoopController.loops">50</stringProp>
```
rantunesboreas: Is there any specific reason why the number of loops here (50) is different than those in the ID fetching thread group (200)?

geneproduct-rest/src/test/jmeter/geneproduct-rest-jmeter-tests.jmx
```diff
@@ -0,0 +1,1365 @@
...
+            <stringProp name="filename"></stringProp>
+          </ResultCollector>
+          <hashTree/>
+        </hashTree>
+      </hashTree>
+      <ThreadGroup guiclass="ThreadGroupGui" testclass="ThreadGroup" testname="Longer search requests" enabled="true">
+        <stringProp name="ThreadGroup.on_sample_error">continue</stringProp>
+        <elementProp name="ThreadGroup.main_controller" elementType="LoopController" guiclass="LoopControlPanel" testclass="LoopController" testname="Loop Controller" enabled="true">
+          <boolProp name="LoopController.continue_forever">false</boolProp>
+          <stringProp name="LoopController.loops">5</stringProp>
```
rantunesboreas: Is there any specific reason why the number of loops here (5) is different than those in the ID fetching thread group (200)?

geneproduct-rest/src/test/jmeter/geneproduct-rest-jmeter-tests.jmx
```diff
@@ -0,0 +1,1365 @@
...
+                <bytes>true</bytes>
+                <threadCounts>true</threadCounts>
+              </value>
+            </objProp>
+            <stringProp name="filename"></stringProp>
+          </ResultCollector>
+          <hashTree/>
+        </hashTree>
+      </hashTree>
+      <ThreadGroup guiclass="ThreadGroupGui" testclass="ThreadGroup" testname="Longer search requests" enabled="true">
```
rantunesboreas: When running the tests I am getting exceptions that should not be occurring:

Here is an example exception:

java.net.URISyntaxException: Illegal character in query at index 83: http://ves-hx-c2.ebi.ac.uk:8081/QuickGO/services/geneproduct/search?query=Bacterial regulatory s, tetR family protein&highlighting=true
	at java.net.URI$Parser.fail(URI.java:2848)
	at java.net.URI$Parser.checkChars(URI.java:3021)
	at java.net.URI$Parser.parseHierarchical(URI.java:3111)
	at java.net.URI$Parser.parse(URI.java:3053)
	at java.net.URI.<init>(URI.java:588)
	at java.net.URL.toURI(URL.java:946)
	at org.apache.jmeter.protocol.http.sampler.HTTPHC4Impl.sample(HTTPHC4Impl.java:252)
	at org.apache.jmeter.protocol.http.sampler.HTTPSamplerProxy.sample(HTTPSamplerProxy.java:74)
	at org.apache.jmeter.protocol.http.sampler.HTTPSamplerBase.sample(HTTPSamplerBase.java:1105)
	at org.apache.jmeter.protocol.http.sampler.HTTPSamplerBase.sample(HTTPSamplerBase.java:1094)
	at org.apache.jmeter.threads.JMeterThread.process_sampler(JMeterThread.java:429)
	at org.apache.jmeter.threads.JMeterThread.run(JMeterThread.java:257)
	at java.lang.Thread.run(Thread.java:745)

geneproduct-rest/src/test/jmeter/geneproduct-rest-jmeter-tests.jmx
```diff
@@ -0,0 +1,1365 @@
...
+            <stringProp name="Argument.name">HOST</stringProp>
+            <stringProp name="Argument.value">${__property(host,,ves-hx-c2.ebi.ac.uk)}</stringProp>
+            <stringProp name="Argument.metadata">=</stringProp>
+          </elementProp>
+          <elementProp name="RAMP" elementType="Argument">
+            <stringProp name="Argument.name">RAMP</stringProp>
+            <stringProp name="Argument.value">${__property(ramp,,10)}</stringProp>
+            <stringProp name="Argument.metadata">=</stringProp>
+          </elementProp>
+          <elementProp name="SIMPLE_QUERY_RESPONSE_TIMEOUT" elementType="Argument">
```
rantunesboreas: I have noticed that there are different timeout lengths specified. Where have the time out values been obtained from? Did the front-end team provide these values to you?

geneproduct-rest/src/test/jmeter/geneproduct-rest-jmeter-tests.jmx
eddturner: The idea was that we have two main use-cases: automated API driven access (e.g., term fetching) and customised user-access (e.g., searching) -- where the former will be accessing data more frequently than the latter. Tony S said a max of 50 concurrent users would be okay, therefore I used this value as the lowest user count, and chose a higher one for the more frequent term fetching (to be on the safe side).

geneproduct-rest/src/test/jmeter/geneproduct-rest-jmeter-tests.jmx
eddturner: 5 is used for the longer running queries, 200 for the shorter ones. This means the number of users ($USERS = 200) each does 5 longer running queries, which is still a lot of longer running queries. However, I wanted to capture the fact long running queries are not the primary use-case we have; typically the most usage will be coming from REST clients driving their interfaces by speaking to our ID lookup endpoints.

geneproduct-rest/src/test/jmeter/geneproduct-rest-jmeter-tests.jmx
eddturner: I did speak to Xav about this yes, so they come from here. However they're configurable via a variable too, so they can be easily changed.

PR #9, opened 2016-05-24
Goa 2089 ontology traversal

eddturner:
This is the framework for traversing the loaded ontologies. Specifically, being able to find the paths / ancestors / descendants related to an ontology term, over a specified set of relationships.

Please just review the differences between this branch, and the pull request "Goa 2092 ontology traversal loading" (i.e., branch matching GOA_2092.*).


ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/OntologyGraph.java
```diff
@@ -0,0 +1,215 @@
...
+    public Set<OntologyRelationship> getEdges() {
+        return ontology.edgeSet();
+    }
+
+    public Set<String> getVertices() {
+        return ontology.vertexSet();
+    }
+
+    public boolean addRelationships(Collection<? extends OntologyRelationship> relationships) {
+        // populate graph with edges, whilst recording the vertices
```
eddturner: Preconditions have been added in downstream branch!

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/OntologyGraph.java
```diff
@@ -0,0 +1,215 @@
...
+                            new OntologyRelationship(oEdge.child, oEdge.parent, oEdge.relationship));
+                }
+        );
+
+        return true;
+    }
+
+    @Override
+    public List<List<OntologyRelationship>> paths(String v1, String v2, OntologyRelationType... relations) {
+        Set<OntologyRelationType> relationsSet = createRelevantRelationsSet(relations);
```
eddturner: Preconditions have been added in downstream branch!

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/OntologyGraph.java
```diff
@@ -0,0 +1,215 @@
...
+        v1v2Paths.removeAll(invalidPaths);
+
+        interestingPaths.addAll(v1v2Paths.stream().map(GraphPath::getEdgeList).collect(Collectors.toList()));
+
+        return interestingPaths;
+    }
+
+    @Override
+    public Set<String> ancestors(String base, OntologyRelationType... relations) {
+        Set<String> ancestorsFound = new HashSet<>();
```
eddturner: Preconditions have been added in downstream branch!

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/OntologyGraph.java
```diff
@@ -0,0 +1,215 @@
...
+                ancestorsFound.add(relation.parent);
+            }
+        }
+
+        return ancestorsFound;
+    }
+
+    @Override
+    public Set<String> descendants(String top, OntologyRelationType... relations) {
+        Set<String> descendantsFound = new HashSet<>();
```
eddturner: Preconditions have been added in downstream branch!

rantunesboreas:
I am not too sure about the migration of the some of the Spring Batch classes into common. 

By moving these, we have basically added Spring Batch libs to all modules that use common, which is all of them!

common/pom.xml
```diff
@@ -30,6 +30,18 @@
             <artifactId>guava</artifactId>
             <version>${guava.version}</version>
         </dependency>
+
+        <dependency>
```
rantunesboreas: Not too sure about adding Spring Batch to common. All modules that import common will also import Spring Batch.

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/OntologyGraphTraversal.java
```diff
@@ -0,0 +1,48 @@
...
+     * Find the list of paths between two vertices in a graph, navigable via
+     * a specified set of relations.
+     *
+     * @param v1 the start vertex from which returned paths must start
+     * @param v2 the end vertex from which returned paths end
+     * @param relations a varargs value containing the relationships over which paths can only travel.
+     *                  By omitting a {@code relation} value, all paths will be returned.
+     * @return a list of paths from {@code child} to {@code parent} via {@code relations}
+     */
+    List<List<OntologyRelationship>> paths(String v1, String v2, OntologyRelationType... relations);
```
rantunesboreas: I'm wondering if we should not represent the vertices as objects. At the moment this solution would only be able to return the node identifier (String), what if we required something else from the node?

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/OntologyRelationType.java
```diff
@@ -0,0 +1,89 @@
...
+    NEGATIVE_REGULATES("NR", "negatively_regulates"),
+    REPLACED_BY("replaced_by", "replaced_by"),
+    CONSIDER("consider", "consider"),
+    HAS_PART("H", "has_part"),
+    OCCURS_IN("OI", "occurs_in"),
+    USED_IN("UI", "used_in"),
+    CAPABLE_OF("CO", "capable_of"),
+    CAPABLE_OF_PART_OF("CP", "capable_of_part_of");
+
+    private static final Map<String, OntologyRelationType> nameToValueMap = new HashMap<>();
```
rantunesboreas: Change name from nameToValueMap to shortNameToValueMap. It makes it easier what name is being mapped.

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/OntologyRelationType.java
```diff
@@ -0,0 +1,89 @@
...
+        return false;
+    }
+
+    /**
+     * Check whether this type is that of a specified type.
+     *
+     * @param type the type to test against
+     * @return true if {@code type} matches this type, or if certain properties are matched. Return false otherwise.
+     */
+    boolean hasType(OntologyRelationType type) {
```
rantunesboreas: This method should be private.

rantunesboreas: The JavaDoc does not explain why the method does checks on so many different combinations of types. Could you add documentation on why the verification of the various types is done this way.

ontology-rest/src/test/java/uk/ac/ebi/quickgo/ontology/traversal/read/StringToOntologyRelationshipMapperTest.java
```diff
@@ -0,0 +1,88 @@
...
+    @Test(expected = IncorrectTokenCountException.class)
+    public void fieldSetWithInsufficientValuesThrowsException() throws Exception {
+        String[] tokens = new String[numColumns() - 1];
+        FieldSet fieldSet = new DefaultFieldSet(tokens);
+
+        mapper.mapFieldSet(fieldSet);
+    }
+
+    @Test
+    public void convertFieldSetWithNullValuesIntoOntologyRelationship() throws Exception {
```
rantunesboreas: Should this actually pass. Should the OntologyRelationship not throw an exception when the values are null?

ontology-rest/src/test/java/uk/ac/ebi/quickgo/ontology/traversal/read/OntologyRelationshipValidatorTest.java
```diff
@@ -0,0 +1,96 @@
...
+    public void findsCorrectNameSpace() {
+        String nameSpace = "nameSpace";
+        String separator = ":";
+        String value = "value";
+        String foundNameSpace = validator.nameSpaceOf(nameSpace + separator + value);
+        assertThat(foundNameSpace, is(nameSpace));
+    }
+
+    @Test(expected = ValidationException.class)
+    public void failsToFindNameSpace() {
```
rantunesboreas: Method should explain why it fails to find the name space, example: nodeWithMissingColonFailsToFindNameSpace()

This way the reader of the test understands that the name space requires a colon somewhere for it to pass.

ontology-rest/src/test/java/uk/ac/ebi/quickgo/ontology/traversal/read/OntologyRelationshipValidatorTest.java
```diff
@@ -0,0 +1,96 @@
...
+        String nameSpace = "nameSpace";
+        String noSeparator = "";
+        String value = "value";
+        validator.nameSpaceOf(nameSpace + noSeparator + value);
+    }
+
+    @Test
+    public void nameSpacesAreValid() {
```
rantunesboreas: maybe a more descriptive name for the test method would help understand what its testing

ontology-rest/src/test/java/uk/ac/ebi/quickgo/ontology/traversal/read/OntologyRelationshipValidatorTest.java
```diff
@@ -0,0 +1,96 @@
...
+        validator.checkValidNameSpaces(relationship);
+    }
+
+    @Test(expected = ValidationException.class)
+    public void differingNameSpacesAreInvalid() {
```
rantunesboreas: differengNameSpacesBetweenParentAndChildVerticesAreInvalid

ontology-rest/src/test/java/uk/ac/ebi/quickgo/ontology/traversal/read/OntologyRelationshipValidatorTest.java
```diff
@@ -0,0 +1,96 @@
...
+        relationship.child = "GO:value";
+        relationship.parent = "ECO:value";
+        validator.checkValidNameSpaces(relationship);
+    }
+
+    @Test(expected = ValidationException.class)
+    public void wrongNameSpacesAreInvalid() {
```
rantunesboreas: I haven't looked at the validator yet, but looking at the test method name I do no understand why the name spaces are wrong.

ontology-rest/src/test/java/uk/ac/ebi/quickgo/ontology/traversal/read/OntologyRelationshipValidatorTest.java
```diff
@@ -0,0 +1,96 @@
...
+        validator.checkValidNameSpaces(relationship);
+    }
+
+    @Test
+    public void relationshipIsValid() {
+        validator.checkValidRelationship(relationship.relationship);
+    }
+
+    @Test(expected = ValidationException.class)
+    public void relationshipIsInvalid() {
```
rantunesboreas: typeOfRelationTypeIsInvalid?

ontology-rest/src/test/java/uk/ac/ebi/quickgo/ontology/traversal/read/OntologyRelationshipValidatorTest.java
eddturner: relationships that go between two namespaces are not valid -- that's why I called the test differingNameSpacesAreInvalid.

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/OntologyGraphTraversal.java
eddturner: True, though the requirement asks for information between vertices involving only vertex names (ids), and paths (lists of relationships -- vertex-relationship-vertex triplets). All other information about a term is accessible from Solr.

If we add more term information (as in legacy QuickGO), we can argue we then move it out of Solr. Then following this trail of though, where does it stop?

eddturner: Another (possibly minor?) point is that more info in the vertex, increases the memory footprint for the graph.

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/OntologyRelationType.java
eddturner: Sure -- this necessary logic comes from legacy QuickGO; will ask TS.

ontology-rest/src/test/java/uk/ac/ebi/quickgo/ontology/traversal/read/StringToOntologyRelationshipMapperTest.java
eddturner: The idea was that the RawOntologyRelationship just captures the raw data in the source file, then the validation step will check it's values are okay or not. I'll check there's a corresponding test for this in validation step.

common/pom.xml
eddturner: This was done because there is batch code used at the indexing + rest component levels. So needed to move reusable batch code somewhere. Want a separate common-batch module ?! The compromise here looks to be, an additional module, or these deps in the existing common.

ontology-rest/src/test/java/uk/ac/ebi/quickgo/ontology/traversal/read/OntologyRelationshipValidatorTest.java
eddturner: yep

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/OntologyGraph.java
```diff
@@ -0,0 +1,215 @@
...
+
+    public Set<OntologyRelationship> getEdges() {
+        return ontology.edgeSet();
+    }
+
+    public Set<String> getVertices() {
+        return ontology.vertexSet();
+    }
+
+    public boolean addRelationships(Collection<? extends OntologyRelationship> relationships) {
```
rantunesboreas: Why does this method return a boolean? Currently the method always returns true.

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/read/OntologyRelationship.java
```diff
@@ -0,0 +1,114 @@
+package uk.ac.ebi.quickgo.ontology.traversal.read;
```
rantunesboreas: This class is part of the model. I don't think it should be placed under the read package, maybe one level up?

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/read/OntologyRelationship.java
```diff
@@ -0,0 +1,114 @@
...
+ * not required by any other class.
+ * <p>
+ * @param <V> the edge type.
+ */
+class LabelledEdge<V> extends DefaultEdge {
+    public V child;
+    public V parent;
+    public String relation;
+
+    LabelledEdge(V child, V v2, OntologyRelationType relation) {
```
rantunesboreas: Rename v2 to parent

rantunesboreas: I would suggest that you add precondition checks to these values. The reason being that most of your graph code expect all fields to be populated, so having a null in any of these fields would cause an NullPointerException later on.

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/read/OntologyRelationship.java
```diff
@@ -0,0 +1,114 @@
...
+        this.child = child;
+        this.parent = v2;
+        this.relation = relation.getShortName();
```
rantunesboreas: Why store this value as a String? Could it not be the enum?

ontology-rest/src/test/java/uk/ac/ebi/quickgo/ontology/traversal/OntologyRelationTypeTest.java
```diff
@@ -0,0 +1,27 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+/**
+ * Validate that ontology relationships are retrieved correctly.
+ *
+ * Created 20/05/16
+ * @author Edd
+ */
+public class OntologyRelationTypeTest {
```
rantunesboreas: This test class does not cover enough usage scenarios contemplated in the OntologyRelationType. Could you add more tests?

ontology-rest/src/test/java/uk/ac/ebi/quickgo/ontology/traversal/read/OntologyGraphConfigTest.java
```diff
@@ -0,0 +1,64 @@
...
+ * {@link OntologyGraph}, by reading test data resources.
+ *
+ * Created 18/05/16
+ * @author Edd
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(
+        classes = {OntologyGraphConfig.class, JobTestRunnerConfig.class},
+        loader = SpringApplicationContextLoader.class)
+public class OntologyGraphConfigTest {
```
rantunesboreas: This should be an integration test, not a unit test. Could you change the name of the file to IT.

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/OntologyGraphTraversal.java
eddturner: I'll do this in the downstream pull request for testing the ontology-rest endpoint if that's okay, as I know I refactored the LabelledEdge class in that one.

PR #11, opened 2016-05-26
Goa 1995 filter by tax

rantunesboreas:
This branch has the code necessary to filter annotations via the taxon id of the geneproduct associated to the annotation.

### DIFF
--- a/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
+++ b/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
@@ -94,7 +94,7 @@ public class AnnotationRequest {
         filters.put(AnnotationFields.TAXON_ID, taxId);
     }
 
-    @Pattern(regexp = "[0-9]+(,[0-9]+)*", message = "Invalid taxonomic identifier: ${validatedValue}")
+    @Pattern(regexp = "[0-9]+(,[0-9]+)*", message = "At least one invalid taxonomic identifier(s): ${validatedValue}")
     public String getTaxon() {
         return filters.get(AnnotationFields.TAXON_ID);
     }
--- a/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
+++ b/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
@@ -75,4 +75,13 @@ public class AnnotationRequest {
     private String[] splitFilterValues(String values) {
         return values.split(COMMA);
     }
+
+    public void setTaxonId(String taxId) {
+        filters.put(AnnotationFields.TAXON_ID, String.valueOf(taxId));
+    }
+
+    @Pattern(regexp = "[0-9]+")
+    public String getTaxonId() {
+        return filters.get(AnnotationFields.TAXON_ID);
+    }
 }
--- a/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
+++ b/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
@@ -77,10 +77,10 @@ public class AnnotationRequest {
     }
 
     public void setTaxon(String taxId) {
-        filters.put(AnnotationFields.TAXON_ID, String.valueOf(taxId));
+        filters.put(AnnotationFields.TAXON_ID, taxId);
     }
 
-    @Pattern(regexp = "[0-9]+")
+    @Pattern(regexp = "[0-9]+(,[0-9]+)*")
     public String getTaxon() {
         return filters.get(AnnotationFields.TAXON_ID);
     }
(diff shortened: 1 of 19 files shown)


annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -89,4 +89,13 @@ public int getPage() {
     private String[] splitFilterValues(String values) {
         return values.split(COMMA);
     }
-}
+
+    public void setTaxon(String taxId) {
+        filters.put(AnnotationFields.TAXON_ID, taxId);
+    }
+
+    @Pattern(regexp = "[0-9]+(,[0-9]+)*")
```
eddturner: If this is only matching a taxon number (e.g., no capturing required), then I this slightly more concise+simpler regex should do the trick:

```java
   @Pattern(regexp = "[0-9]{1,}")
```

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -180,6 +181,61 @@ public void invalidAssignedByThrowsAnError() throws Exception {
...
+
+        response.andExpect(status().isOk())
+                .andExpect(contentTypeToBeJson())
+                .andExpect(totalNumOfResults(2))
+                .andExpect(fieldsInAllResultsExist(2))
+                .andExpect(valuesOccurInField(GENEPRODUCT_ID_FIELD, geneProductId1, geneProductId2));
+    }
+
+    @Test
+    public void invalidTaxIdThrowsError() throws Exception {
```
eddturner: It would be good if the error message in the JSON response was a bit clearer. Currently I see:

```json
{
   "url":"http://localhost/QuickGO/services/annotation/search",
   "message":"The binding of the request parameters to AnnotationRequest uk.ac.ebi.quickgo.annotation.model.AnnotationRequest@2c99c8d has errors, see binding result org.springframework.validation.BeanPropertyBindingResult: 1 errors\nField error in object 'annotationRequest' on field 'taxon': rejected value [-2]; codes [Pattern.annotationRequest.taxon,Pattern.taxon,Pattern.java.lang.String,Pattern]; arguments [org.springframework.context.support.DefaultMessageSourceResolvable: codes [annotationRequest.taxon,taxon]; arguments []; default message [taxon],[Ljavax.validation.constraints.Pattern$Flag;@635ff2a5,[0-9]+(,[0-9]+)*]; default message [must match \"[0-9]+(,[0-9]+)*\"]"}
```

No need to know about binding issues in the error message. I think it'd be fine if it just said that the taxon id is not valid, and should match some specified regex.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -288,6 +344,13 @@ private AnnotationDocument createDocWithAssignedBy(String geneProductId, String
         return doc;
     }
```
eddturner: In `AnnotationRequestValidationIT`, there is a test called `oneValidTaxIdAndOneInvalidTaxIdIsValid`. What happens if a similar request is made in this class? I presume that since `oneValidTaxIdAndOneInvalidTaxIdIsValid` is valid, then the controller will successfully return just the info for the valid taxon. As a tester/user, I'd also expect info about the ids I supplied that were invalid, and the reason why. Is this something we do here?

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
rantunesboreas: You can get several tax ids submitted with a single request, separated by a comma.

eddturner: Cool, in that case, it's fine as is.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
rantunesboreas: Good spot! This required a few changes in the backend.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
rantunesboreas: This was a badly name method. If at least one taxon id is incorrect, then the whole AnnotationRequest fails.

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/ResponseExceptionHandler.java
```diff
@@ -3,14 +3,17 @@
...
+import java.util.stream.Collectors;
 import javax.servlet.http.HttpServletRequest;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.ControllerAdvice;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 
 /**
- * Handler responsible for returning error responses with a meaningful message to the REST client.
+ * Handler responsible for returning error responses with a meaningful messages to the REST client.
```
eddturner: with a meaningful messages -> with meaningful messages

PR #18, opened 2016-06-14
Goa 2101 filter annotations by go evidence code

TonyWardell:


annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -65,6 +65,20 @@ public String getAspect() {
         return filters.get(ASPECT_FIELD);
     }
 
+    /**
+     * The older evidence codes
+     * E.g. IEA, IBA, IBD etc
```
rantunesboreas: It would be nice to have a list, or a link of possible GO evidences in the Javadoc, so that we can refer to it whenever necessary.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -65,6 +65,20 @@ public String getAspect() {
...
+     * @param evidence
+     */
+    public void setGoEvidence(String evidence){
+        filters.put(AnnotationFields.GO_EVIDENCE, evidence);
+    }
+
+    @Pattern(regexp = "^[A-Z]{2,3}(,[A-Z]{2,3})*")
```
rantunesboreas: Can the go evidences not be expressed in lower case?

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/service/converter/AnnotationDocConverterImpl.java
```diff
@@ -34,6 +34,7 @@
             annotation.extensions = new ArrayList<>(annotationDocument.extensions);
         }
 
+        annotation.goEvidence = annotationDocument.goEvidence;
```
rantunesboreas: GO evidence has already been assigned earlier in this block of code. Please remove this one.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequestTest.java
```diff
@@ -60,4 +66,16 @@ public void setAndGetOntologyAspect() {
 
         assertThat(annotationRequest.getAspect(), is(aspect));
     }
-}
\ No newline at end of file
+    @Test
+    public void setAndGetEvidence(){
+        annotationRequest.setGoEvidence(EVIDENCE_IEA);
```
rantunesboreas: Seeing as EVIDENCE_IEA is only used once in this class, could we move the creation down to the method level. I ask this because the class will be testing several input parameters, which means we would have a long list of constants that are only used once or twice.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequestTest.java
```diff
@@ -60,4 +66,16 @@ public void setAndGetOntologyAspect() {
...
+        assertThat(annotationRequest.getGoEvidence(), is(EVIDENCE_IEA));
+    }
+
+    @Test
+    public void setAndGetEvidenceMulti(){
+        annotationRequest.setGoEvidence(EVIDENCE_MULTI);
```
rantunesboreas: Seeing as EVIDENCE_MULTI is only used once in this class, could we move the creation down to the method level. I ask this because the class will be testing several input parameters, which means we would have a long list of constants that are only used once or twice.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequestValidationIT.java
```diff
@@ -65,6 +69,32 @@ public void allAssignedByValuesAreInvalid() {
...
+    public void allGoEvidenceValuesAreValid() {
+        String goEvidenceValues = Arrays.stream(VALID_GO_EVIDENCE)
+                .collect(Collectors.joining(","));
+        annotationRequest.setGoEvidence(goEvidenceValues);
+        assertThat(validator.validate(annotationRequest), hasSize(0));
+    }
+
+    @Test
+    public void allGoEvidenceValuesAreInvalid() {
+        Arrays.stream(INVALID_GO_EVIDENCE).forEach(
```
rantunesboreas: This kind of check is ok to do. However if the test does fail we will not know in which value validation failed. Could you add a logger message to register which value has failed validation. 

Yes, 99% of the times the logger wont be necessary, but for that 1% where validation might fail due to a change in the regex, we will get a clear message for this test.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -53,6 +53,8 @@
 
     private static final String UNAVAILABLE_ASSIGNED_BY = "ZZZZZ";
 
+    private static final String GOEVIDENCE_PARM="goEvidence";
```
rantunesboreas: change to GO_EVIDENCE_PARAM

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -184,6 +186,42 @@ public void invalidAssignedByThrowsAnError() throws Exception {
...
+
+    @Test
+    public void lookupAnnotationFilterByGoEvidenceCodeBySuccessfully() throws Exception {
+        ResultActions response = mockMvc.perform(
+                get(RESOURCE_URL + "/search").param(GOEVIDENCE_PARM, "IEA"));
+
+        response.andExpect(status().isOk())
+                .andExpect(contentTypeToBeJson())
+                .andExpect(totalNumOfResults(genericDocs.size()))
+                .andExpect(fieldsInAllResultsExist(1));
```
rantunesboreas: fieldsInAllResultsExist(1) should not be "1", it should be "genericDocs.size()". If the you put 1 as a method argument only the first result will be checked.

rantunesboreas: It would be good to add an extra check to see if the evidence is in the response of every entry returned.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -184,6 +186,42 @@ public void invalidAssignedByThrowsAnError() throws Exception {
...
+    @Test
+    public void invalidGoEvidenceThrowsException() throws Exception {
+        ResultActions response = mockMvc.perform(
+                get(RESOURCE_URL + "/search").param(GOEVIDENCE_PARM, "BlahBlah"));
+
+        response.andExpect(status().isBadRequest())
+                .andExpect(contentTypeToBeJson());
+
+    }
+
```
rantunesboreas: There is no test for a multi go evidence lookup.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
TonyWardell: We don't need an exhaustive list in the javadoc 1)it won't be to date 2)there is no context

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
TonyWardell: The method to do that is on one of the other filter branches.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
rantunesboreas: That's fine an exhaustive list isn't necessary on the Javadoc, as it will eventually be outdated. However a URL to the most up to date list should be added to the Javadoc. This will provide the reader of the documentation a reference to the existing GO references.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
rantunesboreas: The recommended change to the GOEVIDENCE_PARM was not made.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
rantunesboreas: Could you cherry pick the change from the other branch, and add it to this one?

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
TonyWardell: The jira you wrote says they ids should be considered in a case-insensitive manner. So I've amended the regex and added a test for this.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -53,7 +53,7 @@
 
     private static final String UNAVAILABLE_ASSIGNED_BY = "ZZZZZ";
 
-    private static final String GOEVIDENCE_PARM="goEvidence";
+    private static final String GOEVIDENCE_PARM = "goEvidence";
```
rantunesboreas: I will change GOEVIDENCE_PARM to GOEVIDENCE_PARAM seeing as this has been overlooked.

PR #20, opened 2016-06-14
Add isObsolete to the ontology model.

rantunesboreas:
See issue: https://www.ebi.ac.uk/panda/jira/browse/GOA-2191


eddturner:
Yes all good. Merging.

PR #24, opened 2016-06-24
Goa_2216 Make it possible to create an AllQuery using Negation, and have it serialised correctly to a String

eddturner:


rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/query/SolrQueryConverter.java
```diff
@@ -37,11 +37,16 @@ public SolrQueryConverter(String requestHandler) {
         CompositeQuery.QueryOp operator = query.queryOperator();
         Set<QuickGOQuery> queries = query.queries();
 
-        String operatorText = " " + operator.name() + " ";
+        if (queries.size() == 1 && operator.equals(CompositeQuery.QueryOp.NOT)) {
```
TonyWardell: Swap operands for this line. operator shouldn't be null, but might be null.

PR #28, opened 2016-07-01
Goa 2142 add definition xrefs to ontology model

rantunesboreas:
This branch adds the definition xrefs to the ontology term.

A new definition data-structure within the ontology term has been created to represent the link between the definition itself and the xrefs that support it.


ontology-rest/src/test/java/uk/ac/ebi/quickgo/ontology/service/converter/DefinitionConverterTest.java
```diff
@@ -0,0 +1,123 @@
...
+        doc.definition = text;
+        doc.definitionXrefs = Arrays.asList(xrefText1, xrefText2);
+
+        OBOTerm.Definition def = converter.apply(doc);
+
+        assertThat(def.text, is(text));
+
+        List<OBOTerm.XRef> retrievedXrefs = def.xrefs;
+        assertThat(retrievedXrefs, hasSize(2));
+    }
```
TonyWardell: You should verify the contents of def.xrefs for values as well as size.

rantunesboreas: Done

PR #29, opened 2016-07-15
Goa 1997 filter annotation by gene product type

TonyWardell:
As a QuickGO user
I would like to be able to filter annotations using a specific group to which a set of gene products belong to
So that I can see annotations that are associated to that/those particular gene product(s)
Acceptance criteria:
If the gene product subset does not match any of the ids on the annotation list, then the result should be empty
If the subset matches at least one of the subsets of a gene product in the list of annotations, then the result should only show the affected annotations
The user can input more than one subset, when this is the case then the ids should be ORed.
This filter criteria can be used in conjunction with other non-related filter criteria
The filter should be case-insensitive


annotation-common/src/main/java/uk/ac/ebi/quickgo/annotation/common/document/AnnotationFields.java
```diff
@@ -61,6 +61,7 @@ public static boolean isRetrievable(String field) {
         public static final String REFERENCE_SEARCH = storeAndGet(VALUES, AnnotationFields.REFERENCE_SEARCH);
         public static final String GO_ID = storeAndGet(VALUES, AnnotationFields.GO_ID);
         public static final String GENEPRODUCT_ID = storeAndGet(VALUES, AnnotationFields.GENE_PRODUCT_ID);
+        public static final String GENE_PRODUCT_TYPE = storeAndGet(VALUES, AnnotationFields.DB_OBJECT_TYPE);
```
rantunesboreas: Would it be possible to rename DB_OBJECT_TYPE to GENE_PRODUCT_TYPE?

I think DB_OBJECT_TYPE is too generic, and it conforms to the variable name you came up with.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -40,7 +41,7 @@
 
     private static final String ASPECT_FIELD = "aspect";
     private static final String[] TARGET_FIELDS = new String[]{ASPECT_FIELD, ASSIGNED_BY, TAXON_ID, GO_EVIDENCE,
-            QUALIFIER, REFERENCE_SEARCH, WITH_FROM_SEARCH, ECO_ID, GENE_PRODUCT_ID, GO_ID};
+            QUALIFIER, REFERENCE_SEARCH, WITH_FROM_SEARCH, ECO_ID, GENE_PRODUCT_ID, GO_ID, DB_OBJECT_TYPE};
```
rantunesboreas: Rename DB_OBJECT_TYPE to GENE_PRODUCT_TYPE.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -194,6 +195,17 @@ public String getEcoId(){
         return filterMap.get(ECO_ID);
     }
 
+    public void setGpType(String geneProductType){
+        filterMap.put(DB_OBJECT_TYPE, geneProductType);
+    }
+
+    @Pattern(regexp = "^(?i)(complex|rna|protein)(,(?i)(complex|rna|protein)){0,2}",
```
rantunesboreas: The @Pattern annotation has a flags variable that takes in an array of Flag enum constants. One of those constants is case-insensitivity. This way you don't need to add (?i).

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -890,6 +891,79 @@ public void filterByUnknownReferenceIdIsUnsuccessful() throws Exception {
...
+
+    //----- Tests for GeneProductType ---------------------//
+
+    @Test
+    public void filterByGeneProductTypeReturnsMatchingDocuments() throws Exception {
+
+        ResultActions response = mockMvc.perform(get(RESOURCE_URL + "/search").param(GENE_PRODUCT_TYPE_PARAM,
+                "protein"));
+
+        //Gene Product type is indexed but not stored, so it will not be returned as part of annotation
```
rantunesboreas: This comment is unnecessary, it does not help explain something within the test itself.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -890,6 +891,79 @@ public void filterByUnknownReferenceIdIsUnsuccessful() throws Exception {
...
+        response.andDo(print())
+                .andExpect(status().isOk())
+                .andExpect(contentTypeToBeJson())
+                .andExpect(totalNumOfResults(3))
```
rantunesboreas: Could this be changed to use the NUMBER_OF_GENERIC_DOCS constant.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -890,6 +891,79 @@ public void filterByUnknownReferenceIdIsUnsuccessful() throws Exception {
...
+                .andExpect(fieldsInAllResultsExist(3));
+    }
+
+
+    @Test
+    public void filterByGeneProductTypesSingleParameterReturnsMatchingDocuments() throws Exception {
```
rantunesboreas: The test method name does not reflect what the test is doing. Could you either change the name, or the test itself?

rantunesboreas: looks like the test below does the multiple params match, so I guess its the test itself that is not reflecting the test method's intention.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -890,6 +891,79 @@ public void filterByUnknownReferenceIdIsUnsuccessful() throws Exception {
...
+        repository.save(doc);
+
+
+        ResultActions response = mockMvc.perform(get(RESOURCE_URL + "/search").param(GENE_PRODUCT_TYPE_PARAM,"rna"));
+
+        response.andDo(print())
+                .andExpect(status().isOk())
+                .andExpect(contentTypeToBeJson())
+                .andExpect(totalNumOfResults(0))
+                .andExpect(fieldsInAllResultsExist(0));
```
rantunesboreas: This check is unnecessary, seeing as no results were returned.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequestValidationIT.java
```diff
@@ -369,6 +369,40 @@ public void ecoIdIsInvalid() {
         );
     }
 
+    //GENE PRODUCT TYPE PARAMETER
+    @Test
+    public void validGeneProductTypeValuesDontCauseAnError(){
```
rantunesboreas: Test method refers to a gene product "type", however the test has three of them "complex,rna,protein". Could you change the method name, or the test itself to reflect the right conditions.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequestValidationIT.java
```diff
@@ -369,6 +369,40 @@ public void ecoIdIsInvalid() {
...
+                    annotationRequest.setGpType(invalidValue);
+                    Set<ConstraintViolation<AnnotationRequest>> violations = validator.validate(annotationRequest);
+                    assertThat(violations, hasSize(is(1)));
+                    assertThat(violations.iterator().next().getMessage(),
+                            is("At least one 'Gene Product Type' value is invalid: " + invalidValue));
+                }
+        );
+    }
+
+
```
rantunesboreas: Please remove extra white spaces.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequestValidationIT.java
```diff
@@ -369,6 +369,40 @@ public void ecoIdIsInvalid() {
...
+    @Test
+    public void validGeneProductTypeValuesDontCauseAnError(){
+        String validIds = "complex,rna,protein";
+        annotationRequest.setGpType(validIds);
+        assertThat(validator.validate(annotationRequest), hasSize(0));
+    }
+
+
+    @Test
+    public void validGeneProductTypeNotCaseSensitive(){
```
rantunesboreas: Same as previous comment.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -194,6 +195,17 @@ public String getEcoId(){
         return filterMap.get(ECO_ID);
     }
 
+    public void setGpType(String geneProductType){
+        filterMap.put(DB_OBJECT_TYPE, geneProductType);
```
rantunesboreas: This can create a potential bug! The reason being that if geneProductType is mixed case or upper case, then you will get 0 results back.

Looking at the schema.xml for annotation:
<field name="dbObjectType" type="string" indexed="true" stored="false" multiValued="false"/>

you see that the type is string.

I would probably suggest "filterMap.put(DB_OBJECT_TYPE, geneProductType.toLowerCase());" so as to avoid lowercasing it in solr. If we opt for lowercasing it at the solr level, we can't use that field to facet upon.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
TonyWardell: It does explain why we are not checking the results for our search value, which we do in 
Nevertheless I've removed it.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
TonyWardell: Added a test and renamed tests to clarify testing.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequestValidationIT.java
TonyWardell: I think the important word in that method name is valueS

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequestValidationIT.java
TonyWardell: Changed emphasis from passed in values to operation of the method.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequestValidationIT.java
rantunesboreas: Your right, must have skipped the values bit when I was reading the method name.

PR #31, opened 2016-07-18
Goa 2245 make gene product filter case insensitive

TonyWardell:
All searching / filtering should be case-insensitive; to not do this would be user-hostile. Use the CASE_INSENSITIVE flag when constructing the Pattern object.


eddturner:
I think it would be useful to the caller if they could specify whether or not they want case-sensitive matching. It depends on the use-case, e.g., :
* input from a client might want to be forgiving, so having case-insensitive matching as okay
* input from a flat-file, e.g., if this were to be used during indexing, or from an expert client, would not want to be case-insensitive

PR #32, opened 2016-07-21
GOA-2018_annotation_statistics_goId

rantunesboreas:
Merge branch 'master' into GOA-2018_annotation_statistics_goId

# Conflicts:
#	annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/AnnotationREST.java


rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/AggregateFunction.java
```diff
@@ -0,0 +1,31 @@
+package uk.ac.ebi.quickgo.rest.search;
+
+/**
+ * Enumerates the aggregation functions supported by the domain.
+ *
```
TonyWardell: Would this be clearer with an example in the javadoc?

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/service/statistics/StatisticsService.java
```diff
@@ -0,0 +1,23 @@
...
+import uk.ac.ebi.quickgo.annotation.model.StatisticsGroup;
+import uk.ac.ebi.quickgo.rest.search.results.QueryResult;
+
+/**
+ * Contract for a service that provides annotation based statistics.
+ *
+ * @author Ricardo Antunes
+ */
+public interface StatisticsService {
+
```
TonyWardell: Interesting one this.
Search service vs statistics service (verb vs noun)
One is a generic service (Search ) vs specific service (here's the thing we are returning)
Maybe statistics is a specialized form of calculation service <Statistics> verb<thing we are calculating>
Just another way of looking at why things are called what they are

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/results/Aggregation.java
```diff
@@ -0,0 +1,153 @@
...
+import uk.ac.ebi.quickgo.rest.search.AggregateFunction;
+import uk.ac.ebi.quickgo.rest.search.query.Aggregate;
+
+import com.google.common.base.Preconditions;
+import java.util.LinkedHashSet;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Stores the result of the aggregation section found within the response of the underlying data source.
```
TonyWardell: This javadoc line is pretty unclear and relies on the reader knowing quite a bit more than what is here. Could it be clearer, have examples?

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/solr/SolrResponseAggregationConverter.java
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    private double convertToDouble(Object number) {
+        double convertedValue;
+
+        if (number instanceof Double) {
+            convertedValue = (double) number;
+        } else if (number instanceof Integer) {
+            convertedValue = (double) (int) number;
```
TonyWardell: Nicer to have convertedValue = ((Integer)number).doubleValue() ?

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/query/Aggregate.java
```diff
@@ -0,0 +1,84 @@
...
+import uk.ac.ebi.quickgo.rest.search.AggregateFunction;
+
+import com.google.common.base.Preconditions;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Data-source representing which supported aggregation fields {@link AggregateFunction} should be executed over
+ * the provided field.
```
TonyWardell: This javadoc could explain in a bit more detail - for instance why have nested aggregate? Would an example help?

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/solr/SolrResponseAggregationConverter.java
```diff
@@ -0,0 +1,166 @@
...
+
+                aggregation.addAggregationResult(function, name, hits);
+            }
+        } else if (BUCKETS_ID.equals(field)) {
+            List<NamedList<?>> buckets = (List<NamedList<?>>) value;
+            buckets.forEach(bucket -> convertBucket((NamedList<?>) bucket, aggregation));
+        } else {
+            logger.debug("Did not process field: {} with value: {}", field, value);
+        }
+    }
```
TonyWardell: This method is very busy with the 'what-to-do'  line by line. 
Its also suffering with lots of if/else clauses.
Could it be made clearer by extracting methods that hide the detail but explain what each of the clauses are for and why are the doing what they do?

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/service/statistics/AnnotationStatisticsService.java
```diff
@@ -0,0 +1,164 @@
...
+    private QueryRequest buildQueryRequest(AnnotationRequest request) {
+        BasicSearchQueryTemplate basicTemplate = new BasicSearchQueryTemplate(Collections.emptyList());
+        AggregateSearchQueryTemplate aggregateTemplate = new AggregateSearchQueryTemplate();
+
+        BasicSearchQueryTemplate.Builder basicBuilder = basicTemplate.newBuilder()
+                .setQuery(QuickGOQuery.createAllQuery())
+                .setFilters(request.createFilterRequests().stream()
+                        .map(converterFactory::convert)
+                        .collect(Collectors.toSet()))
+                .setPage(FIRST_PAGE)
```
TonyWardell: Does BasicSearchQueryTemplate need to deal with pages - is it a requirement of the conversion to a solr query? As pages are relevant here.
What is the relationship between DefaultSearchQueryTemplate and BasicSearchQueryTemplate atm.

BTW its very similar to the code in AnnotationController.

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/query/Aggregate.java
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * @author Ricardo Antunes
+ */
+public class Aggregate {
```
TonyWardell: I can see that the design is (to) aggregate -> creates aggregation 
But an aggregate is also something that has be aggregated, not about to be 
I think it would be clearer if aggregate was called AggregationRequest or something like it to avoid confusion.
I think its purpose be clearer.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/service/statistics/StatisticsService.java
rantunesboreas: Are you suggesting we rename the interface?

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/results/Aggregation.java
rantunesboreas: Done

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/solr/SolrResponseAggregationConverter.java
rantunesboreas: Done

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/query/Aggregate.java
rantunesboreas: Done

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/solr/SolrResponseAggregationConverter.java
rantunesboreas: I have tried to simplify the code, and rename the methods to not be as confusing.

As for the if/else clauses, these exist because solrJ has not data model to represent aggregations. The data exists within a very basic data structure called NamedList. This data structure mixes the different hierarchies into one big object. Hence the excessive checks to make it workl.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/service/statistics/AnnotationStatisticsService.java
rantunesboreas: Currently BasicSearchQueryTemplate deals with paging, it is a basic requirement for all our query templates, so might as well code it in the BasicQueryTemplate.
There is no relationship between the two, DefaultSearchQueryTemplate was the initial implementation, it needs to be eventually replaced by the new query templates.

Two more query templates will be created in the future: FacetedQuerySearchTemplate, and HighlightingQuerySearchTemplate. Depending on the requirements of the response, any of these could be used, or they can be composed.

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/query/Aggregate.java
rantunesboreas: Renamed Aggregate to AggregateRequest and Aggregation to AggregateResponse.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/service/statistics/StatisticsService.java
TonyWardell: I'm wondering if the Statistics service can be replaced with using SearchService with the right generic, and then the form to get the data for /stats would mirror that for a normal search.
For stats the query request is built inside of the stats service. For normal search its done within the controller. It would be nice if the same arrangement could be used in both cases, with stats be a different version of search (which in a respect it is) rather than a different case (almost entirely)

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/results/Aggregation.java
TonyWardell: Much more comprehensive.

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/query/Aggregate.java
TonyWardell: >If the aggregate is based on a name, then the name is name of the name

Do you mean if the aggregate is based on a name, then the name is the name?

Can the javadoc be reformatted to be within column 80.

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/solr/SolrResponseAggregationConverter.java
TonyWardell: I think the changes could go further.

` private void convertAggregateValue(String field, Object value, AggregateResponse aggregation) {
        String function = SolrAggregationHelper.determineFunction(field);

        if (isFieldWithFunction(function) {
            if (isAggregatedFunction(function)) {
				populateAggregation(field, value, aggregation)
            } else {
                addAggregationFunctionToAggregation(prefix, field, value, aggregation);
            }
        } else if (BUCKETS_ID.equals(field)) {
            List<NamedList<?>> buckets = (List<NamedList<?>>) value;
            buckets.forEach(bucket -> convertBucket((NamedList<?>) bucket, aggregation));
        } else {
            logger.debug("Did not process field: {} with value: {}", field, value);
        }
    }
	
	private boolean isFieldWithFunction(String function){
			return !function.isEmpty()
	}
	
	private boolean isAggregatedFunction(String function){
			return (AGG_TYPE_PREFIX.equals(function))
	}
	
	private boolean populateAggregation(String field, Object value, AggregateResponse aggregation){
		AggregateResponse nestedAggregation = createNestedAggregation(SolrAggregationHelper.nameOfAFunctionalValue(field), value);
        aggregation.addNestedAggregation(nestedAggregation);
	}`

TonyWardell: So I only changed it up to else if (BUCKETS..)

but the point is to say why having prefix is or isn't important and to make it clear what prefix represents.

SolrAggregationHelper.fieldPrefixExtractor(field); method has not javadoc so you can't glean that information from there.

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/query/Aggregate.java
TonyWardell: Coolio

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/query/Aggregate.java
rantunesboreas: The Javadoc was changed due to a refactor that I did. I did not pickup that the refactor had changed the wording within the Javadoc. I have changed the wording to make it understandable.

As for the reformatting, the UniProt coding standards state that the length of a line is 120, this applies to Javadoc too.

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/solr/SolrResponseAggregationConverter.java
rantunesboreas: Have added a few more methods and Javadoc to make the code easier to understand.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/service/statistics/StatisticsService.java
rantunesboreas: I'm not to sure about this. The search service is used for returning search results form a user. Whilst the Statistics service runs analytics on a search result. The stats service is not interested in returning search results, in fact it doesn't return search results at all. What it does return are a set of distribution stats made on certain fields for a given user query. 

If you want we can discuss this further to see what can be done to improve the readability.

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/query/Aggregate.java
TonyWardell: The EOL marker for my intellij shows a vertical line at col. 120 (I thought it was at 80) , so I'm afraid it's too wide.

PR #33, opened 2016-07-22
Goa 2265 ontology rest ancestors for obsolete terms

eddturner:


ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/service/OntologyServiceImpl.java
```diff
@@ -69,7 +75,9 @@ private OntologyServiceImpl() {}
     }
 
     @Override public List<T> findCompleteInfoByOntologyId(List<String> ids) {
-        return convertDocs(ontologyRepository.findCompleteByTermId(ontologyType, buildIdList(ids)));
+        return convertDocsToCoreData(ontologyRepository.findCompleteByTermId(ontologyType, buildIdList(ids)))
```
rantunesboreas: It would be nice if we could use convertDocs() in this instance. After looking at the code convertDocs() uses convertDocsToCoreData(). So in the spirit of uniformity having all find methods call convertDocs would make it easier to read.

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/service/OntologyServiceImpl.java
```diff
@@ -102,40 +110,65 @@ private OntologyServiceImpl() {}
     }
 
     @Override public List<T> findAncestorsInfoByOntologyId(List<String> ids, OntologyRelationType... relations) {
-        return convertDocs(ontologyRepository.findCoreAttrByTermId(ontologyType, buildIdList(ids)))
-                .stream()
-                .map(term -> {
-                    term.ancestors = ontologyTraversal.ancestors(singleton(term.id), relations);
-                    return term;
-                })
+        return convertDocsWithoutTraversalData(ontologyRepository.findCoreAttrByTermId(ontologyType, buildIdList(ids)))
```
rantunesboreas: couldn't this be done by just calling convertDocs()?

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/service/OntologyServiceImpl.java
```diff
@@ -102,40 +110,65 @@ private OntologyServiceImpl() {}
...
     }
 
+    List<String> buildIdList(List<String> ids) {
+        Preconditions.checkArgument(ids != null, "List of IDs cannot be null");
+
+        return ids.stream().map(queryStringSanitizer::sanitize).collect(Collectors.toList());
+    }
+
     List<T> convertDocs(List<OntologyDocument> docs) {
+        return convertDocsToCoreData(docs).collect(Collectors.toList());
```
rantunesboreas: Could we rename this method. The issue is that we already have a definition of core in the Ontology service, and that definition is not the same as the core that this method references to.

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/service/OntologyServiceImpl.java
```diff
@@ -102,40 +110,65 @@ private OntologyServiceImpl() {}
...
+    }
+
+    private Stream<T> convertDocsWithoutTraversalData(List<OntologyDocument> docs) {
```
rantunesboreas: I would suggest renaming this method. to convertDocs() and omit the "withoutTraversalData". I would then create a method which adds ancestors or descendants to the existing stream of converted docs. What do you think?

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/service/OntologyServiceImpl.java
```diff
@@ -102,40 +110,65 @@ private OntologyServiceImpl() {}
...
         return docs.stream()
-                .map(converter::convert)
-                .map(this::insertAncestors)
-                .collect(Collectors.toList());
+                .map(converter::convert);
     }
 
-    List<String> buildIdList(List<String> ids) {
-        Preconditions.checkArgument(ids != null, "List of IDs cannot be null");
+    private Stream<T> convertDocsToCoreData(List<OntologyDocument> docs) {
```
rantunesboreas: Rename method name.

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/service/OntologyServiceImpl.java
```diff
@@ -102,40 +110,65 @@ private OntologyServiceImpl() {}
...
+            ancestors = Collections.unmodifiableList(Collections.emptyList());
+        }
+        term.ancestors = ancestors;
+
+        return term;
     }
 
-    private int calculatePageNumber(int oneBasedPageNum) {
-        return oneBasedPageNum - 1;
+    private T insertDescendants(T term, OntologyRelationType... relations) {
```
rantunesboreas: This code is very similar to the ancestors, is there any way we could avoid the code duplication?

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/service/OntologyServiceImpl.java
eddturner: I ended up leaving this name but providing javadocs. Hopefully the shape of the rest of the class is clearer now?

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/service/OntologyServiceImpl.java
eddturner: Code duplication removed, so less to maintain, however, note that the alternative solution isn't much smaller! Since it's more functional now, it should read a bit more clearly though.

PR #34, opened 2016-07-22
GOA-2144_Add_replaces_to_ontology_model

rantunesboreas:


rantunesboreas:
I have merged the replacements branch into this one, seeing as there were code changes that lead to a reuse of code for both branches.

PR #35, opened 2016-07-25
GOA-2251_Add_replacements_to_ontology

rantunesboreas:


indexing/src/main/java/uk/ac/ebi/quickgo/index/ontology/converter/GenericTermToODocConverter.java
```diff
@@ -87,10 +86,13 @@
         return null;
     }
 
-    protected List<String> extractConsidersAsList(GenericTerm term) {
-        if (!isEmpty(term.consider())) {
-            return term.consider().stream()
-                    .map(GenericTerm::getId)
+    protected List<String> extractRelations(Collection<TermRelation> replaceList) {
```
eddturner: should this be called `extractReplacementTypes` or something?

indexing/src/test/java/uk/ac/ebi/quickgo/index/ontology/converter/GenericTermToODocConverterTest.java
```diff
@@ -35,25 +34,91 @@ public void setup() {
 
     // considers
     @Test
-    public void extractsNoConsideredWhenNull() {
-        when(term.consider()).thenReturn(null);
+    public void extractsNoRelationsWhenNull() {
```
eddturner: If method is renamed above (`extractRelations` -> `extractReplacementTypes` or something similar), then this test name can be changed too. Just a reminder. Just needs to be in sync.

ontology-common/src/test/java/uk/ac/ebi/quickgo/ontology/common/OntologyRepositoryIT.java
```diff
@@ -105,9 +105,26 @@ public void retrieves2DocsCoreFieldsOnly() {
     }
```
eddturner: probably originating from previous changes, but looks like this class could have its methods reordered according to visibility rules (intellij ctrl-shift-a, then type "rearrange", then <enter>). ... I noticed a private method higher than a public test method.

eddturner:
Changes look good. Other than the comments I've made explicitly in code, I think the main thing is the name of `relations` throughout -- should this be `replacementTypes` or something similar? I'm not sure why it's `relations`.

eddturner:
btw -- looks like master is broken for now, so this shouldn't be merged until that's fixed.

indexing/src/main/java/uk/ac/ebi/quickgo/index/ontology/converter/GenericTermToODocConverter.java
rantunesboreas: Have created methods to make it easier to understand when to extract a replacement and when to extract a replaces.

indexing/src/test/java/uk/ac/ebi/quickgo/index/ontology/converter/GenericTermToODocConverterTest.java
rantunesboreas: extractRelation is used in extractReplacements and extractReplaces. Code is used in both case, hence the name being a bit more generic.

rantunesboreas:
I have made your requested changes.

I have also decided to remove the replacedBy attribute from OBOTerm and OntologyDocument. seeing as it is already represented in the replacements section.

eddturner:
cool. 

Just double checked tests before merging and i see there's a failure after a renamed variable.
E.g., see
`uk.ac.ebi.quickgo.index.ontology.converter.GenericTermToODocConverterTest#convertSimpleFields`

PR #36, opened 2016-07-26
Add case sensitivity flag for Gene Product ID validation

TonyWardell:
As discussed.

### DIFF
--- a/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/service/search/SearchServiceConfig.java
+++ b/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/service/search/SearchServiceConfig.java
@@ -43,7 +43,9 @@ public class SearchServiceConfig {
     @Value("${geneproduct.db.xref.valid.regexes}")
     String xrefValidationRegexFile;
 
-    @Value("${geneproduct.db.xref.valid.casesensitive}")
+    private static final boolean DEFAULT_XREF_VALIDATION_IS_CASE_SENSITIVE = true;
+
+    @Value("${geneproduct.db.xref.valid.casesensitive:"+DEFAULT_XREF_VALIDATION_IS_CASE_SENSITIVE+"}")
     boolean xrefValidationCaseSensitive;
 
     private static final String COMMA = ",";
--- a/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/service/search/SearchServiceConfig.java
+++ b/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/service/search/SearchServiceConfig.java
@@ -43,6 +43,9 @@ public class SearchServiceConfig {
     @Value("${geneproduct.db.xref.valid.regexes}")
     String xrefValidationRegexFile;
 
+    @Value("${geneproduct.db.xref.valid.casesensitive}")
+    boolean xrefValidationCaseSensitive;
+
     private static final String COMMA = ",";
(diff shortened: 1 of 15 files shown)


eddturner:
Doesn't compile -- looks like `DbXRefEntityTest` needs updating.

annotation-rest/src/test/resources/search.properties
```diff
@@ -4,3 +4,5 @@ search.return.fields=id,geneProductId,qualifier,goId,goEvidence,ecoId,reference,
 search.field.repo2domain.map=
 search.highlight.delims=
 geneproduct.db.xref.valid.regexes=src/test/resources/DB_XREFS_ENTITIES.dat.gz
+## =============== Validation  =================
+geneproduct.db.xref.valid.casesensitive=false
```
eddturner: i'd recommend adding this also to:
`annotation-rest/src/main/resources/search.properties` so it's clear this is required

geneproduct-rest/src/test/java/uk/ac/ebi/quickgo/geneproduct/loader/DbXRefLoaderTest.java
```diff
@@ -22,14 +22,14 @@
 
 	@Test
 	public void loadFileUnsuccessfully(){
-		DbXRefLoader dbXRefLoader = new DbXRefLoader(NOWHERE_CANTFIND);
+		DbXRefLoader dbXRefLoader = new DbXRefLoader(NOWHERE_CANTFIND, false);
```
eddturner: small one -- would be worth changing `false` to a constant that says `CASE_INSENSITIVE_MATCHING` to help readability. and same for other tests. And `true` to a constant, `CASE_SENSITIVE_MATCHING`. Ditto then for the other tests.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/service/search/SearchServiceConfig.java
```diff
@@ -43,6 +43,9 @@
     @Value("${geneproduct.db.xref.valid.regexes}")
     String xrefValidationRegexFile;
 
+    @Value("${geneproduct.db.xref.valid.casesensitive}")
```
eddturner: Would make sense to have a default value for whether case sensitivity is used.

```java
// something along the lines of the following:
...
private static final boolean DEFAULT_XREF_VALIDATION_IS_CASE_SENSITIVE = true;
...
@Value("${geneproduct.db.xref.valid.casesensitive:"+DEFAULT_XREF_VALIDATION_IS_CASE_SENSITIVE+"}")
```

... then if for some reason, this property is not provided in the properties file, then we'll get the default value.

TonyWardell: Got point. Done.

geneproduct-rest/src/test/java/uk/ac/ebi/quickgo/geneproduct/loader/DbXRefLoaderTest.java
TonyWardell: CASE_INSENSITIVE_MATCHING = false
makes it sound like the argument is turning in-sensitivity off.

how do you feel about 
CASE_SENSITIVE_MATCHING  = false;

eddturner: yep 'tis fine, good good

PR #37, opened 2016-07-27
Add target sets (BHF-UCL, KRUK) etc to Annotation to be indexed.

TonyWardell:


indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
```diff
@@ -94,7 +92,15 @@ private int extractTaxonId(String rawTaxonId) {
         return value == null ? null : Arrays.asList(value.split(PIPE));
     }
 
+    private List<String> createNullableListFromCommaSeparatedValue(String value) {
```
rantunesboreas: This method looks very similar to createNullableListFromPipeSeparatedValue(), could you try to refactor the code so that we have a single converter method?

annotation-common/src/main/java/uk/ac/ebi/quickgo/annotation/common/document/AnnotationFields.java
```diff
@@ -30,6 +30,7 @@
     public static final String GENE_PRODUCT_TYPE = "geneProductType";
     public static final String DB_SUBSET = "dbSubset";
     public static final String TAXON_ID = "taxonId";
+    public static final String TARGET_SET = "targetSet";
```
rantunesboreas: Seeing as its a searchable field we should add it to the Searchable inner class.

rantunesboreas:
One question, can the search for the target sets be case insensitive?

annotation-common/src/main/java/uk/ac/ebi/quickgo/annotation/common/document/AnnotationFields.java
TonyWardell: I've done that in the filtering branch.

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
TonyWardell: Yes - I thought about this original but felt passing another parameter around was messier than stating the intent with duplicated (ish) line. I've changed it to suggested.

PR #38, opened 2016-07-27
GOA-2177_Add_planned_changes_to_ontology_model

rantunesboreas:


indexing/src/test/java/uk/ac/ebi/quickgo/index/ontology/converter/GOTermToODocConverterTest.java
```diff
@@ -191,4 +194,46 @@ public void extractsBlacklistWhenNotExists() {
...
+
+        String title2 = "signalling";
+        String url2 = "http://wiki.geneontology.org/index.php/Signaling";
+
+        GOTerm.NamedURL plannedChange2 = new GOTerm.NamedURL(title2, url2);
+
+        when(term.getPlannedChanges()).thenReturn(Arrays.asList(plannedChange1, plannedChange2));
+
+        List<String> expectedDiscussions = converter.extractGoDiscussions(term);
+        assertThat(expectedDiscussions, hasSize(2));
```
eddturner: a bit tedious i know, but probably should make sure the contents of the `expectedDiscussions` is what we expect. That would def be enough tho

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/controller/OBOController.java
```diff
@@ -47,13 +47,28 @@
  * @author Edd
  */
 public abstract class OBOController<T extends OBOTerm> {
+    static final String TERMS_RESOURCE = "terms";
```
eddturner: nice ... thanks for the refactoring -- clearer now

PR #39, opened 2016-07-28
GOA-2237_Add_credits_to_ontology_model

rantunesboreas:
Contains code to add credits (acknowledgments to projects that help fund the creation of the ontology term).

### DIFF
--- a/ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/service/converter/GoDiscussionConverter.java
+++ b/ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/service/converter/GoDiscussionConverter.java
@@ -22,7 +22,7 @@ import static uk.ac.ebi.quickgo.common.converter.FlatFieldBuilder.newFlatField;
  * <p>
  * @author Ricardo Antunes
  */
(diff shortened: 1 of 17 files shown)


ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/service/converter/GoDiscussionConverter.java
```diff
@@ -22,7 +22,7 @@
  * <p>
  * @author Ricardo Antunes
  */
-public class GoDiscussionConverter implements FieldConverter<GOTerm.GoDiscussion> {
+class GoDiscussionConverter implements FieldConverter<GOTerm.GoDiscussion> {
```
eddturner: I must have missed this last time (apologies) but a very minor point would be a suggestion to rename `GoDiscussion...` to `GODiscussion...`, since "gene ontology" are two words, and using java's naming convention, would make this "GO"...

I think this affects `GoDiscussion`, `GoDiscussionConverter` and `GoDiscussionConverterTest` as far as I can see.

eddturner:
only 1 minor comment to address -- renaming a few `Go.*` classes to `GO.*`.

PR #40, opened 2016-07-29
Goa 2241 regex for rna not working

TonyWardell:
I've added a test for case sensitivity for RNA, Intact and UniProt gene product ids. When loading the validation regexes in from file they work fine, which is why the jira request for the branch is closed, but this test proves this.

### DIFF
--- a/annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
+++ b/annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
@@ -70,9 +70,6 @@ public class AnnotationControllerIT {
     private static final String ECO_ID2 = "ECO:0000323";
     private static final String NOTEXISTS_ECO_ID3 = "ECO:0000888";
 
-    //Configuration
-    private static final int NUMBER_OF_GENERIC_DOCS = 3;
-
     private MockMvc mockMvc;
     private List<AnnotationDocument> genericDocs;
     @Autowired
@@ -88,8 +85,8 @@ public class AnnotationControllerIT {
         mockMvc = MockMvcBuilders.
                 webAppContextSetup(webApplicationContext)
                 .build();
-
-        genericDocs = createGenericDocs(NUMBER_OF_GENERIC_DOCS);
+        final int numberOfGenericDocs = 3;
+        genericDocs = createGenericDocs(numberOfGenericDocs);
         repository.save(genericDocs);
     }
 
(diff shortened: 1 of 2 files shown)


annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -400,7 +400,32 @@ public void filterByUniProtKBAndIntactAndRNACentralGeneProductIDSuccessfully() t
         response.andExpect(status().isOk())
                 .andExpect(contentTypeToBeJson())
                 .andExpect(totalNumOfResults(NUMBER_OF_GENERIC_DOCS))
-                .andExpect(fieldsInAllResultsExist(NUMBER_OF_GENERIC_DOCS));
+                .andExpect(fieldsInAllResultsExist(NUMBER_OF_GENERIC_DOCS))
```
rantunesboreas: NUMBER_OF_GENERIC_DOCS should not be used here. Reason being that this value represents the number of documents stored in the repository before the test starts.

It just happens to match the number of documents that are retrieved (3). I would suggest creating a value within the test method, e.g. expectedNumberOfDocs = 3

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -400,7 +400,32 @@ public void filterByUniProtKBAndIntactAndRNACentralGeneProductIDSuccessfully() t
...
+        repository.save(AnnotationDocMocker.createAnnotationDoc(intactGp));
+        repository.save(AnnotationDocMocker.createAnnotationDoc(rnaGp));
+        StringJoiner sj = new StringJoiner(",");
+        sj.add(uniprotGp).add(intactGp).add(rnaGp);
+        ResultActions response = mockMvc.perform(
+                get(RESOURCE_URL + "/search").param(GP_PARAM, sj.toString()));
+
+        response.andExpect(status().isOk())
+                .andExpect(contentTypeToBeJson())
+                .andExpect(totalNumOfResults(NUMBER_OF_GENERIC_DOCS))
```
rantunesboreas: Same comment as above.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
TonyWardell: I've in-lined all values. For many tests the expectedNumberOfDocs  just happens to be one of a number of numeric values to be tested, and the validation method names make intent clear.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
TonyWardell: As above.

TonyWardell: So code is back as it was, apart from the two changes you've requested. Creating the required annotation with the select-by data only in each test method would not work without removing the tests for completeness of model data, so I think they should remain in.

PR #44, opened 2016-08-04
GOA-2135_Redirect_secondary_terms

rantunesboreas:
Allows a client to lookup a GO or ECO term by using its secondary identifier.


eddturner:
All good.

PR #49, opened 2016-08-05
Goa 2022 stats by gp taxon

rantunesboreas:


annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerStatisticsIT.java
```diff
@@ -147,14 +131,76 @@ public void ontologyStatsForFilteredDocsContaining2OntologyIdsReturns2OntologyId
...
+    public void statsForAllDocsContaining1TaxonIdReturns1TaxonIdStat() throws Exception {
+        Set<String> savedTaxonIds = selectValuesFromDocs(savedDocs, doc -> String.valueOf(doc.taxonId));
+        assertThat(savedTaxonIds, hasSize(1));
+
+        String type = AnnotationFields.TAXON_ID;
+
+        ResultActions response = mockMvc.perform(get(STATS_ENDPOINT));
+
+        assertStatsResponse(response, type, NUMBER_OF_GENERIC_DOCS, savedTaxonIds);
+    }
```
TonyWardell: The json in the statistics response starts with 

`{
  "numberOfHits": 2,
  "results": [
    {
      "groupName": "annotation",
      "totalHits": 6,
      "types": [
        {
          "type": "taxonId",
          "values": [
            {
              "key": "12345",
              "percentage": 1,
              "hits": 6
            }
          ]
        },
        {
          "type": "goId",
    `

I presume the two hits are the two groupnames (annotation and geneproduct) (rather than the two types of facets within (taxonid and goid). If so - can numberOfHits become numberOfDataSets (or something better). This information isn't required as far as I know so should it be dropped from the results?

TonyWardell: The `assertStatsResponse` method doesn't check if the calculations have been preformed correctly. Can it?

rantunesboreas: Number of hits corresponds to the number of objects returned by the results array. This behaviour is consistent with all of the other endpoints we have in our services. Changing it from numberOfHits to numberOfGroups could be done, however I don't see a major gain in changing this.

rantunesboreas: The assertStatsResponse does not check the actual numbers. We do have unit tests that check if the values are being read correctly. 

By introducing these checks we would just be confirming that Solr is collecting the correct values for the facets, which I think we can safely assume.

PR #55, opened 2016-08-09
Goa 2023 add assigned by to stats calc

rantunesboreas:


eddturner:
I find test failures for `AnnotationControllerIT` and the other `AnnotationController*IT` tests.

eddturner:
Other than the test failures, the code looks good and is following the same pattern as the other stats components.

rantunesboreas:
Fixed broken tests.

PR #56, opened 2016-08-09
Goa 2273 filter annotations by target set

TonyWardell:
A set of gene products can be defined for a project. This set is the gene products the project is interested in annotating.
QuickGO must allow annotations to be filtered by the gene products sets defined (aka target sets)

### DIFF
--- a/annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
+++ b/annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
@@ -7,6 +7,7 @@ import uk.ac.ebi.quickgo.annotation.common.document.AnnotationDocument;
 import uk.ac.ebi.quickgo.annotation.service.search.SearchServiceConfig;
 import uk.ac.ebi.quickgo.common.solr.TemporarySolrDataStore;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.StringJoiner;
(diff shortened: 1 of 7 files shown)


annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -1020,52 +1019,42 @@ public void filterBySingleGeneProductSubsetReturnsDocuments() throws Exception {
...
-                .andExpect(totalNumOfResults(NUMBER_OF_GENERIC_DOCS + 1))
-                .andExpect(fieldsInAllResultsExist(NUMBER_OF_GENERIC_DOCS + 1));
+                .andExpect(totalNumOfResults(NUMBER_OF_GENERIC_DOCS))
+                .andExpect(fieldsInAllResultsExist(NUMBER_OF_GENERIC_DOCS));
     }
 
-
     @Test
-    public void filterByMultipleGeneProductSubsetInMultipleParametersReturnsDocuments() throws Exception {
+    public void filterByTwoTargetSetValuesReturnsMatchingDocuments() throws Exception {
```
rantunesboreas: Instead of checking if the alternate targetSet hits all documents, could you create a document with a target set that is not present in the rest of the documents, and check if the filtering works properly?

TonyWardell: Done.

PR #60, opened 2016-08-17
GOA-2311_Refactor ontology controller to use new search templates

rantunesboreas:
Refactor ontology controller to use OntologyRequest and BasicSearchTemplate


ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/model/OntologyRequest.java
```diff
@@ -0,0 +1,61 @@
...
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+
+import static uk.ac.ebi.quickgo.rest.search.DefaultSearchQueryTemplate.DEFAULT_PAGE_NUMBER;
+
+/**
+ * A data structure used to store the input parameters a client can submit to the Ontology search enpoint
+ *
+ * Once the comma separated values have been set, then turn then into an object (SimpleFilter) that
```
eddturner: then turn then -> then turn them

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/controller/ECOController.java
```diff
@@ -32,9 +31,8 @@
     @Autowired
     public ECOController(OntologyService<ECOTerm> ecoOntologyService,
             SearchService<OBOTerm> ontologySearchService,
-            SearchableField searchableField,
```
eddturner: Given that this value is now effectively replaced by the call to `ontologyRetrievalConfig.getSearchReturnedFields()`, is there still a need for the `@Component` `OntologyFieldSpec` ?

eddturner: ... Hmm, looking at this replacement, I'm not sure this'll have the same behaviour.

In `uk.ac.ebi.quickgo.ontology.service.search.SearchServiceConfig.OntologyCompositeRetrievalConfig#getSearchReturnedFields`, we see:

```java
           @Override public List<String> getSearchReturnedFields() {
                return Arrays.asList(ontologySearchSolrReturnedFields.split(COMMA));
            }
```

And the value this method is the value now replacing the previous `SearchableField searchableField`.

Originally, (i.e., the `SearchableField`) provided access to all the fields we've identified as searchable (indexed Solr fields) in the repository. But it's now been replaced by the field names that are specified in the properties file, that are to be returned from the response (stored Solr fields).

Was this replacement intentional?

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/model/OntologyRequest.java
rantunesboreas: Done

PR #73, opened 2016-09-12
Goa 2321 presets for references

eddturner:


quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/presets/read/PresetsConfig.java
```diff
@@ -22,7 +23,7 @@
  * @author Edd
  */
 @Configuration
-@Import({PresetsCommonConfig.class, AssignedByPresetsConfig.class})
+@Import({PresetsCommonConfig.class, AssignedByPresetsConfig.class, ReferencePresetsConfig.class})
```
rantunesboreas: Wouldn't it be better to use @ComponentScan instead of having to add each preset individually?

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/presets/read/PresetsConfig.java
```diff
@@ -32,9 +33,13 @@
     @Bean
     public Job presetsBuildJob(
             JobBuilderFactory jobBuilderFactory,
-            Step assignedByStep) {
+            Step assignedByStep,
+            Step referenceGenericDbStep,
+            Step referenceSpecificDbStep) {
         return jobBuilderFactory.get(PRESET_LOADING_JOB_NAME)
                 .start(assignedByStep)
+                .next(referenceGenericDbStep)
```
rantunesboreas: It would be great if we could somehow register these steps dynamically.

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/presets/read/PresetsConfig.java
rantunesboreas: In PresetsCommonConfig there is a @ComponentScan({"uk.ac.ebi.quickgo.rest"}), is it necessary?

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/presets/read/reference/ReferencePresetsConfig.java
```diff
@@ -0,0 +1,124 @@
...
+    public static final String CORE_REFERENCE_DB_LOADING_STEP_NAME = "GenericReferenceDBReadingStep";
+    public static final String SPECIFIC_REFERENCE_LOADING_STEP_NAME = "SpecificReferenceReadingStep";
+    private static final String REFERENCE_DB_DEFAULTS = "DOI,GO_REF,PMID,REACTOME";
+    private static final String REFERENCE_SPECIFIC_DB_DEFAULTS =
+            "0000037,0000038,0000043,0000045,0000039,0000040,0000044,0000046,0000003,0000041,0000020,0000002," +
+                    "0000042,0000019,0000035,0000049,0000104,0000108";
+    private static final RawNamedPreset INVALID_PRESET = null;
+    private static final String GO_REF_FORMAT = "GO_REF:%s";
+
+    @Value("#{'${reference.db.preset.source:}'.split(',')}")
```
rantunesboreas: If we were to use yml files, this splitting would no longer be necessary

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/presets/read/reference/ReferencePresetsConfig.java
```diff
@@ -0,0 +1,124 @@
...
+                .build();
+    }
+
+    private String buildGORefID(String name) {
+        return String.format(GO_REF_FORMAT, name);
+    }
+
+    private ItemProcessor<RawNamedPreset, RawNamedPreset> rawPresetFilter(Set<String> validPresetNames) {
+        return rawNamedPreset ->
+                validPresetNames.contains(rawNamedPreset.name) ? rawNamedPreset : INVALID_PRESET;
```
rantunesboreas: Isn't the checking of presence of a name already being done by the validator? Should we need to check again?

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/presets/read/reference/ReferencePresetsConfig.java
```diff
@@ -0,0 +1,124 @@
...
+
+        return stepBuilderFactory.get(SPECIFIC_REFERENCE_LOADING_STEP_NAME)
+                .<RawNamedPreset, RawNamedPreset>chunk(chunkSize)
+                .faultTolerant()
+                .skipLimit(SKIP_LIMIT)
+                .<RawNamedPreset>reader(rawPresetMultiFileReader(specificResources, itemReader))
+                .processor(compositeItemProcessor(
+                        rawPresetValidator(),
+                        rawPresetFilter(specificDBDefaults)))
+                .writer(rawItemList -> rawItemList.forEach(rawItem -> {
```
rantunesboreas: This is a bit hard to follow, could you put the writing logic into a method?

rantunesboreas:
Could we change the package name where the loading logic for the presets is found, from presets.read to  service.loader.preset or something similar?

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/presets/read/ff/RawNamedPresetValidator.java
```diff
@@ -0,0 +1,26 @@
...
+ * Validates the {@link RawNamedPreset} instances being read from flat-files and mapped via
+ * {@link StringToRawNamedPresetMapper}.
+ *
+ * Created 31/08/16
+ * @author Edd
+ */
+public class RawNamedPresetValidator implements ItemProcessor<RawNamedPreset, RawNamedPreset> {
+    @Override public RawNamedPreset process(RawNamedPreset rawNamedPreset) throws Exception {
+        if (rawNamedPreset == null) {
+            throw new ValidationException("RawDBPreset cannot be null or empty");
```
rantunesboreas: Should this not be RawNamedPreset?

quickgo-client/src/test/java/uk/ac/ebi/quickgo/client/model/presets/PresetItemTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Test(expected = IllegalArgumentException.class)
+    public void nullDescriptionCausesExceptionIn2ParamConstructor() {
+        new PresetItem(VALID_NAME, null);
+    }
+
+    @Test
+    public void canCreateValidPreset() {
+        PresetItem presetItem = new PresetItem(VALID_NAME, VALID_DESCRIPTION, VALID_RELEVANCY);
+        assertThat(presetItem, is(notNullValue()));
```
rantunesboreas: could you check that the values are populated correctly.

quickgo-client/src/test/java/uk/ac/ebi/quickgo/client/presets/read/PresetsFailedRelevancyFetchingIT.java
```diff
@@ -44,7 +45,8 @@
     public void loadDefaultAssignedByPresetsAfterFailedRESTInfoFetching() throws Exception {
         assertThat(preset.assignedBy.getPresets(), hasSize(0));
 
-        JobExecution jobExecution = jobLauncherTestUtils.launchJob();
+        JobExecution jobExecution =
+                jobLauncherTestUtils.launchStep(AssignedByPresetsConfig.ASSIGNED_BY_LOADING_STEP_NAME);
         BatchStatus status = jobExecution.getStatus();
 
         assertThat(status, is(BatchStatus.COMPLETED));
```
rantunesboreas: for the second assertion I would suggest two things:
   1 - Move the extraction of the preset name list into a method that clearly indicates the intention
   2 - Make ASSIGNED_BY_DEFAULTS = "UniProtKB" in the AssignedByPresetsConfig class public, so that you can access it in this test. It makes the readbility much clearer

quickgo-client/src/test/java/uk/ac/ebi/quickgo/client/presets/read/PresetsFailedRelevancyFetchingIT.java
```diff
@@ -44,7 +45,8 @@
     public void loadDefaultAssignedByPresetsAfterFailedRESTInfoFetching() throws Exception {
```
rantunesboreas: Hmm maybe this test should be moved to where the relevancy is checked (Assigned by), probably not a good idea to have a central one, because things could get confusing

quickgo-client/src/test/java/uk/ac/ebi/quickgo/client/presets/read/PresetsSuccessfulRelevancyFetchingIT.java
```diff
@@ -20,17 +23,15 @@
...
+import static uk.ac.ebi.quickgo.client.presets.read.MockPresetDataConfig.*;
 
 /**
  * Created 31/08/16
  * @author Edd
  */
 @RunWith(SpringJUnit4ClassRunner.class)
 @ContextConfiguration(
-        classes = {PresetsConfig.class, MockRESTConfig.class, JobTestRunnerConfig.class},
+        classes = {PresetsConfig.class, MockPresetDataConfig.class, JobTestRunnerConfig.class},
```
rantunesboreas: Probably best to split these into different classes, there will be more presets, and I can see this class grow, and get confusing.

PR #76, opened 2016-09-15
GOA-2344_Create_annotation_stress_tests

rantunesboreas:


eddturner:
There could/should be a test for gene-product subsets, and withFrom?

rantunesboreas:
Have enabled the aspect test.

solr-cores/src/main/cores/annotation/conf/schema.xml
```diff
@@ -106,8 +106,8 @@
     <field name="symbol_unsorted" type="lowercase" indexed="true" stored="false" multiValued="false"/>
     <copyField source="symbol" dest="symbol_unsorted"/>
 
-    <field name="dbSubset_unsorted" type="lowercase" indexed="true" stored="false" multiValued="false"/>
-    <copyField source="dbSubset" dest="dbSubset_unsorted"/>
+    <field name="geneProductSubset_unsorted" type="lowercase" indexed="true" stored="false" multiValued="false"/>
```
eddturner: Just noticed that this change should also be reflected in any `search.properties` and `application.properties` which refer to fields. E.g., `annotation.terms.query.compatible.fields` property (and possibly `search.return.fields`?)

rantunesboreas: Done

PR #84, opened 2016-10-05
Goa 2323 presets for taxon, qualifier, aspects, geneproducttype

eddturner:
PR for the smaller preset components


quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/model/presets/impl/CompositePresetImpl.java
```diff
@@ -153,6 +180,64 @@ private void ifPresetItemMatchesThenApply(
...
+        GO_SLIMS_SETS,
+        TAXONS,
+        QUALIFIERS,
+        ASPECTS
+    }
+
+    private static class StaticAspects {
+
+        static final String MOLECULAR_FUNCTION = "Molecular Function";
+        static final String FUNCTION = "function";
```
rantunesboreas: We use the canonical form represent the aspects. It used to be the short form, but we changed it because the canonical is the de facto representation.

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/model/presets/impl/CompositePresetImpl.java
```diff
@@ -153,6 +180,64 @@ private void ifPresetItemMatchesThenApply(
...
         GENE_PRODUCT,
-        GO_SLIMS_SETS
+        GENE_PRODUCT_TYPES,
+        GO_SLIMS_SETS,
+        TAXONS,
+        QUALIFIERS,
+        ASPECTS
+    }
+
+    private static class StaticAspects {
```
rantunesboreas: Why not model this as an enum?

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/model/presets/impl/CompositePresetImpl.java
```diff
@@ -153,6 +180,64 @@ private void ifPresetItemMatchesThenApply(
...
+                            .withId(PROCESS).build());
+            presetAspects.add(
+                    PresetItem
+                            .createWithName(CELLULAR_COMPONENT)
+                            .withId(COMPONENT).build());
+            return presetAspects;
+        }
+    }
+
+    private static class StaticGeneProductTypes {
```
rantunesboreas: Why not model this as an enum?

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/PresetsConfigHelper.java
```diff
@@ -94,4 +102,31 @@ static JobExecutionListener logJobListener() {
...
+
+    public static ItemReader<RawNamedPreset> topItemsFromRESTReader(
+            RESTFilterConverterFactory converterFactory,
+            String field) {
+        FilterRequest request = FilterRequest.newBuilder().addProperty(field).build();
+
+        try {
+            List<String> relevantItems = converterFactory.<List<String>>convert(request).getConvertedValue();
+            Iterator<String> relevantItemIterator = relevantItems.iterator();
+            AtomicInteger position = new AtomicInteger(0);
```
rantunesboreas: Why do you need an AtomicInteger?

eddturner: just because it's being used in a lambda, so the value needs to be effectively final. If using an int/Integer, incrementing inside the lambda will complain because if it is final, then we can't update it's value.

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/model/presets/impl/CompositePresetImpl.java
eddturner: done

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/model/presets/impl/CompositePresetImpl.java
eddturner: done -- yeh, bit shorter now.

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/model/presets/impl/CompositePresetImpl.java
eddturner: Do you need any changes for this? These two values are just the ones the front-end will display. What they do with it I'm not 100% sure of ...

rantunesboreas: Ok

PR #85, opened 2016-10-06
introduce new chart sub-resource to contain all coordinate layouting information

eddturner:
introduce new chart sub-resource to contain all coordinate layouting information of the terms within the image chart seen at the "chart" sub-resource.

### DIFF
--- a/ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/controller/OBOController.java
+++ b/ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/controller/OBOController.java
@@ -1,5 +1,6 @@
 package uk.ac.ebi.quickgo.ontology.controller;
 
+import uk.ac.ebi.quickgo.graphics.model.GraphImageLayout;
 import uk.ac.ebi.quickgo.graphics.ontology.RenderingGraphException;
 import uk.ac.ebi.quickgo.graphics.service.GraphImageService;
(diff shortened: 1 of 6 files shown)


ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/controller/OBOController.java
```diff
@@ -374,6 +376,31 @@ public OBOController(OntologyService<T> ontologyService,
...
+    public ResponseEntity<GraphImageLayout> getChartCoordinates(@PathVariable(value = "ids") String ids) {
+        try {
+            GraphImageLayout layout = graphImageService
+                    .createChart(validationHelper.validateCSVIds(ids), getOntologyType().name()).getLayout();
+
+            return ResponseEntity
+                    .ok()
+                    .body(layout);
+        } catch (RenderingGraphException e) {
+            throw new RetrievalException(e);
```
rantunesboreas: Hmmm, we need to talk about this. I'm not sure this is the best way to report this exception. By re-throwing the exception we are indirectly revealing internals of the service to the user. I think we should be logging the original exception as an error, and throwing a new user friendly exception. Hopefully one that does reveal internals.

eddturner: logging and throwing a different one is fine by me. no worries.

PR #88, opened 2016-10-12
GOA-2395 separating presets response so that different components can be retrieved on demand

eddturner:
Allow the user to declare which preset fields they want returned in the response. use jackson's simplefilter mechanism to achieve this.


quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/controller/PresetsController.java
```diff
@@ -35,12 +41,53 @@ public PresetsController(CompositePreset presets) {
...
      *
+     * @param fields the preset fields wanted. If empty, all fields are returned
      * @return a populated instance that encapsulates the preset information
      */
     @ApiOperation(value = "Provides preset filtering information indicating valid terms and a corresponding " +
             "description; all of which are ordered by relevancy.")
     @RequestMapping(method = {RequestMethod.GET}, produces = {MediaType.APPLICATION_JSON_VALUE})
-    public ResponseEntity<CompositePreset> compositePreset() {
-        return ResponseEntity.ok(presets);
+    public FilteredCompositePreset compositePreset(@RequestParam(name = "fields", required = false) String fields) {
```
rantunesboreas: Why not receive fields as a String[] instead of String. This way you wont need to treat them as csvs?

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/controller/PresetsController.java
```diff
@@ -35,12 +41,53 @@ public PresetsController(CompositePreset presets) {
...
+            checkArgument(compositePreset != null, "CompositePreset cannot be null");
+            this.setFilters(filterNothing());
+        }
+
+        private static SimpleFilterProvider filterNothing() {
+            return new SimpleFilterProvider().addFilter(COMPOSITE_PRESET_FILTER, SimpleBeanPropertyFilter.serializeAllExcept());
+        }
+
+        void showFields(String csvFields) {
+            Set<String> fieldsSet = Stream.of(csvFields.split(COMMA)).collect(Collectors.toSet());
```
rantunesboreas: No need for this if you change the endpoint to accept a String[] or a List<String>

PR #90, opened 2016-10-19
GOA-2398_make_taxon_name_indexible

rantunesboreas:


eddturner:
`GeneProductRepositoryIT` doesn't pass

geneproduct-rest/src/test/java/uk/ac/ebi/quickgo/geneproduct/search/GeneProductUserQueryScoringIT.java
```diff
@@ -635,9 +1170,12 @@ public void partialInSameWordReturnsShortestMatchesFirst() throws Exception {
     public void partialInDifferentWordsReturnsShortestMatchesFirst() throws Exception {
```
eddturner: Cool. Looks like all combinations tested.

Note for future. I wonder if we could reduce our tests here and base them all on transitivity of precedence. E.g.,

(high to low)
```
1.) symbol exact
2.) taxon exact
3.) synonym exact
4.) name exact
5.) symbol word case insensitive
6.) taxon word case insensitive
7.) synonym/name word case insensitive
8.) symbol substring
9.) taxon substring
10.) synonym/name substring
```

Our tests are pretty comprehensive, testing all permutations as far as I can see. I suppose it should be fine to reduce the tests to just:

1 vs 2
2 vs 3
3 vs 4

etc., and by transitivity of scoring, it'd be possible to deduce 1 vs 4 is also tested implicitly.

What do you think? Am I missing something? No need to do anything now, just wanted to make a comment.

rantunesboreas: Yes, we could do this, I was just full proofing things.

PR #96, opened 2016-10-25
sort annotation search results by geneproduct id always

eddturner:


rantunesboreas:
Will this sorting work in conjuction with {!terms} queries?

eddturner:
Tested and it seems to work fine.

http://ves-hx-cf:8080/QuickGO/services/annotation/search?usage=descendants&goId=GO:0016020 (1k+ descendants ~3s)
http://ves-hx-cf:8080/QuickGO/services/annotation/search?usage=descendants&goId=GO:0005575 (5k+ descendants ~5s)

I checked the solr logs and as wanted, a terms query was being issued to solr.

PR #111, opened 2016-11-11
Goa 2192 validate qualifiers annotation filtering

TonyWardell:


annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -300,6 +300,7 @@ public void setQualifier(String... qualifier) {
         filterMap.put(QUALIFIER, qualifier);
     }
 
+    @ArrayPattern(regexp = "^[A-Za-z_]+$", paramName = QUALIFIER)
```
rantunesboreas: How is the NOT represented represented in this regex? As far as I understood all positive qualifiers can be negated. Examples: NOT | enables; NOT | involved_in; NOT | part_of

TonyWardell: Fixed.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -300,7 +300,7 @@ public void setQualifier(String... qualifier) {
         filterMap.put(QUALIFIER, qualifier);
     }
 
-    @ArrayPattern(regexp = "^[A-Za-z_]+$", paramName = QUALIFIER)
+    @ArrayPattern(regexp = "^(?i:NOT \\| )?[A-Za-z_]+$", paramName = QUALIFIER)
```
TonyWardell: Why not indeed! - changed as suggested.

PR #114, opened 2016-12-02
GOA-2468 limit the number of pages one can paginate through.

eddturner:
It has been set to a maximum of 25. This will encourage search refinement and prevent deep paging performance issues which affect often search engines in general.


annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -456,7 +457,8 @@ public void setLimit(int limit) {
         this.limit = limit;
     }
 
-    @Min(value = MIN_PAGE_NUMBER, message = "Page size cannot be less than {value} but found: ${validatedValue}")
+    @Min(value = MIN_PAGE_NUMBER, message = "Page number cannot be less than {value}, but found: ${validatedValue}")
```
rantunesboreas: If you are referencing the MAX_PAGE_NUMBER in the ControllerValidationHelper, then it makes sense to move the MIN_PAGE_NUMBER over to the same class.

It seems like the ContollerValidationHelper is a central class for validation. If this is the case, why not move the following constants over as well: DEFAULT_ENTRIES_PER_PAGE, MAX_ENTRIES_PER_PAGE, MIN_ENTRIES_PER_PAGE?

eddturner: ah this is done

PR #120, opened 2016-12-09
Goa 2545 fix co terms not being written

TonyWardell:
A copy and paste error meant the CoTermsAll data was being written to the CoTermsManual file.

### DIFF
--- a/indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/coterms/CoTermsConfig.java
+++ b/indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/coterms/CoTermsConfig.java
@@ -46,10 +46,10 @@ public class CoTermsConfig {
             "together", "compared"};
     private static final String DELIMITER = "\t";
 
-    @Value("${indexing.coterms.manual:/nfs/public/rw/goa/quickgo_origin/full/CoTermsManual}")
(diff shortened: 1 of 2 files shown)


indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/coterms/CoTermsConfig.java
```diff
@@ -49,7 +49,7 @@
     @Value("${indexing.coterms.manual:/nfs/public/rw/goa/quickgo_origin/full/CoTermsManual}")
     private String manualCoTermsPath;
 
-    @Value("${indexing.coterms.manual:/nfs/public/rw/goa/quickgo_origin/full/CoTermsAll}")
+    @Value("${indexing.coterms.all:/nfs/public/rw/goa/quickgo_origin/full/CoTermsAll}")
```
rantunesboreas: I do not agree with providing the default value, its too specific. If you want a default value then is should be a directory that is relative to the location where the software is being executed.

rantunesboreas: Same thing for manual.

TonyWardell: Changed to use a subdir of the user's home dir, rather than plonk it somewhere random.

PR #124, opened 2016-12-20
Goa 2511 number of hits for coterms

TonyWardell:
Number of hits is now the total number of Co-occurring terms for a record, not the number the client asked for or is limited too.

### DIFF
--- a/ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/controller/CoTermController.java
+++ b/ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/controller/CoTermController.java
@@ -78,7 +78,7 @@ public class CoTermController {
         validateGoTerm(id);
 
         final List<CoTerm> coTerms = coTermRepository.findCoTerms(id, toCoTermSource(source));
-        return getResultsResponse(coTerms.stream()
+        return getResultsResponse(coTerms.size(), coTerms.stream()
                 .filter(ct -> ct.getSimilarityRatio() >= similarityThreshold)
                 .limit(workoutLimit(limit))
                 .collect(Collectors.toList()));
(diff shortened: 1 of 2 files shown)


ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/controller/CoTermController.java
```diff
@@ -90,11 +90,14 @@ public CoTermController(CoTermRepository coTermRepository) {
      * @param results a list of results
      * @return a {@link ResponseEntity} containing a {@link QueryResult} for a list of results
      */
-    private <ResponseType> ResponseEntity<QueryResult<ResponseType>> getResultsResponse(List<ResponseType> results) {
+    private <ResponseType> ResponseEntity<QueryResult<ResponseType>> getResultsResponse(long
```
eddturner: can we change add a new line after the '('?

so from : `getResultsReponse(long`

to 
```
getResultsResponse(
      long numberOfHits,
      List<ResponseType> results)
```

ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/controller/CoTermController.java
```diff
@@ -90,11 +90,14 @@ public CoTermController(CoTermRepository coTermRepository) {
...
+            numberOfHits, List<ResponseType>
+            results) {
         if (results == null) {
```
eddturner: not a big deal, but sonar reminded me that we should introduce a new variable for `results` -- rather than assigning a value to a method param.

e.g., 
```java
List<ResponseType> queryResults;
if (results == null) {
    queryResults = Collections.emptyList();
}
... // and use queryResults in place of results from here on

PR #125, opened 2016-12-20
Change gene product id validation

TonyWardell:
Change gene product id validation to cope with ids structured in the format id:v…ariation or feature.


common/src/main/java/uk/ac/ebi/quickgo/common/validator/DbXRefEntity.java
```diff
@@ -48,7 +48,14 @@ public DbXRefEntity(
     /**
      * This method checks whether the supplied {@code id} matches the regular expression
```
eddturner: this class could do with a formatting (just saw a couple of lines needing indenting)

PR #128, opened 2017-01-04
Goa 2552 annotations in consistent order

TonyWardell:


annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -86,6 +86,29 @@ public void setup() {
         repository.save(genericDocs);
     }
 
+    // CONSISTENT ORDER
+    @Test
+    public void annotationsAlwaysReturnedInOrderWrittenToRepository() throws Exception{
+
+        String geneProductId = "A0";
```
eddturner: I think the intent of this test would be clearer if we added say 3 documents, rather than just 1, such as:

* add doc whose GP id is BBBBB
* add doc whose GP id is ZZZZZ
* add doc whose GP id is AAAAA

then the test can expect these three to be in the order they were added, and it'll be clear it has nothing to do with GP id sorting.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -86,6 +86,29 @@ public void setup() {
...
+                .andExpect(contentTypeToBeJson())
+                .andExpect(totalNumOfResults(4))
+                .andExpect(fieldInRowHasValue(GENEPRODUCT_ID_FIELD, 0, genericDocs.get(0).geneProductId))
+                .andExpect(fieldInRowHasValue(GENEPRODUCT_ID_FIELD, 1, genericDocs.get(1).geneProductId))
+                .andExpect(fieldInRowHasValue(GENEPRODUCT_ID_FIELD, 2, genericDocs.get(2).geneProductId))
+                .andExpect(fieldInRowHasValue(GENEPRODUCT_ID_FIELD, 3, geneProductId));
+    }
+
+
     //ASSIGNED BY
```
eddturner: Is `uk.ac.ebi.quickgo.annotation.controller.AnnotationControllerIT#resultsReturnedInOrderWrittenToSolrRelevancyNotUsed` needed? Is it covered by the test above? If yes, then this can be removed.

If needed, the name of the test could be clearer. Change to something like `resultsReturnedInOrderWrittenToSolr` -- the test  code should make it clear through variable naming, etc., that relevancy is not being used.

I think the line comments within this test method are a bit confusing, and needn't be there, and the test would be clearer if the documents needed by the expected results are added explicitly in the test method.

The last line of:

```java
response.andDo(print())
                .andExpect(status().isOk())
                .andExpect(contentTypeToBeJson())
                .andExpect(totalNumOfResults(4))
                .andExpect(fieldInRowHasValue("geneProductId", 0, expected));
```

Shouldn't all geneProductIds be checked?

solr-cores/src/main/cores/annotation/conf/solrconfig.xml
```diff
@@ -897,7 +897,6 @@
             <str name="wt">json</str>
             <str name="indent">true</str>
             <str name="df">text</str>
-            <str name="sort">score desc, geneProductId asc</str>
```
eddturner: As chatted about last year, I think it's unfortunately going to be necessary to include a new field, `rowNumber` or something, which we order by explicitly in `solrconfig.xml`, because we're using a cloud setup and we can't guarantee that the results will come back from the different nodes in the order in which they were added.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
TonyWardell: Removed.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
TonyWardell: Changed.

TonyWardell:
Requested changes made.

solr-cores/src/main/cores/annotation/conf/schema.xml
```diff
@@ -148,6 +148,10 @@
     <field name="extension_search" type="extension_ignorecase" indexed="true" stored="false" multiValued="true"/>
     <copyField source="extension" dest="extension_search"/>
 
+    <!-- ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ -->
+    <!-- Fields that are used specifically for Ordering -->
+    <field name="rowNumber" type="row_number" indexed="true" stored="true" multiValued="false"/>
```
eddturner: this field should be a required field

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -86,6 +83,33 @@ public void setup() {
...
+        String geneProductId2 = "ZZZZZ";
+        String geneProductId3 = "AAAAA";
+
+        List<AnnotationDocument> documents = new ArrayList<>();
+        documents.add(AnnotationDocMocker.createAnnotationDoc(geneProductId1));
+        documents.add(AnnotationDocMocker.createAnnotationDoc(geneProductId2));
+        documents.add(AnnotationDocMocker.createAnnotationDoc(geneProductId3));
+        repository.save(documents);
+
+        ResultActions response = mockMvc.perform(
```
eddturner: When I debugged this and checked the contents of the repository, I see exceptions:

```java
org.springframework.core.convert.ConversionFailedException: Failed to convert from type [java.lang.String] to type [long] for value 'AAAAA!202280816411190'; nested exception is java.lang.NumberFormatException: For input string: "AAAAA!202280816411190"
	at org.springframework.core.convert.support.ConversionUtils.invokeConverter(ConversionUtils.java:41)
	at org.springframework.core.convert.support.GenericConversionService.convert(GenericConversionService.java:192)
	at org.springframework.core.convert.support.GenericConversionService.convert(GenericConversionService.java:176)
```

This looks to be related to the `rowNumber` field, which we don't need in `AnnotationDocument`. So, hopefully, removing `rowNumber` from that class will mean this exception doesn't occur.

annotation-common/src/main/java/uk/ac/ebi/quickgo/annotation/common/AnnotationDocument.java
```diff
@@ -13,6 +13,7 @@
  * @author Edd
  */
 public class AnnotationDocument implements QuickGODocument {
+    public long rowNumber;
```
eddturner: as discussed, don't need

annotation-common/src/main/java/uk/ac/ebi/quickgo/annotation/common/AnnotationDocument.java
```diff
@@ -68,6 +69,7 @@
     @Field(AnnotationFields.DATE)
     public Date date;
 
+
```
eddturner: remove extra line

annotation-common/src/test/java/uk/ac/ebi/quickgo/annotation/common/document/AnnotationDocMocker.java
```diff
@@ -59,7 +59,8 @@ public static AnnotationDocument createAnnotationDoc(String geneProductId) {
         // automatically compute a document identifier,
         // to overcome non-uniqueness of all other annotation fields
         // (in solrconfig.xml this is set automatically as a UUID)
-        doc.id = geneProductId + "-" + System.nanoTime();
+        doc.rowNumber = System.nanoTime();
```
eddturner: no longer needed

annotation-common/src/test/java/uk/ac/ebi/quickgo/annotation/common/document/AnnotationDocMocker.java
```diff
@@ -59,7 +59,8 @@ public static AnnotationDocument createAnnotationDoc(String geneProductId) {
...
+        doc.id = geneProductId + "!" + doc.rowNumber;
```
eddturner: doc.rowNumber -> nanotime value

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -86,6 +83,33 @@ public void setup() {
...
+                get(RESOURCE_URL + "/search"));
+
+        response.andDo(print())
+                .andExpect(status().isOk())
+                .andExpect(contentTypeToBeJson())
+                .andExpect(totalNumOfResults(3))
+                .andExpect(fieldInRowHasValue(GENEPRODUCT_ID_FIELD, 0, geneProductId1))
+                .andExpect(fieldInRowHasValue(GENEPRODUCT_ID_FIELD, 1, geneProductId2))
+                .andExpect(fieldInRowHasValue(GENEPRODUCT_ID_FIELD, 2, geneProductId3));
```
eddturner: a few extra expectations as discussed just to check:

`geneProductId1.id`'s row number component < 
`geneProductId2.id`'s row number component <
`geneProductId3.id`'s row number component

This is a bit pedantic I know, but I do feel it's necessaary; testing here is slightly different to testing searchable fields.

* our searchable field tests test hits when we expect them (and the order of results returned is defined by the "row  number" / "insertion order".
* if we check here that the row number components are ordered, then we know the sort order in the requesthandler (in solrconfig.xml) worked
   * ... and by showing the requesthandler did what it should do, then we have by-passed the limitation of testing on a single node solr instance -- the results we see in this test should apply to a distributed solrcloud search too.

 Moreover, this test underpins all search result orderings for all annotation queries.

solr-cores/src/main/cores/annotation/conf/schema.xml
eddturner: And it needn't be stored just to confirm as chatted about

PR #132, opened 2017-01-17
Goa 2579 cursor mark in solr request and response

eddturner:


eddturner:
I will ask Boris to review this, to help distribute the load

rest-common/src/test/java/uk/ac/ebi/quickgo/rest/search/query/QueryRequestTest.java
```diff
@@ -32,42 +42,38 @@ public void buildsQueryRequestOnlyWithQuery() throws Exception {
...
-
-        new QueryRequest.Builder(query).setPageParameters(-1, 2).build();
+        new QueryRequest.Builder(query).setPage(new RegularPage(-1, 2)).build();
     }
 
     @Test
-    public void buildsQueryRequestWithQueryAndPageParameterComponents() throws Exception {
-        QuickGOQuery query = QuickGOQuery.createQuery("field1", "value1");
-
+    public void buildsQueryRequestWithQueryAndRegularPageParameterComponents() throws Exception {
```
borkabu: Indeed, I agree with Sonar that it is too generic to throw Exception. However, in this case throw Exception clause not needed anywhere in this file. Everything runs without it.

rest-common/src/test/java/uk/ac/ebi/quickgo/rest/search/results/QueryResultTest.java
```diff
@@ -80,7 +73,11 @@ public void validQueryResultWithNoPageInfoAndNoFacet() throws Exception {
     public void validFullQueryResult() throws Exception {
         long numberOfHits = 2;
         List<String> results = Arrays.asList("result1", "result2");
-        uk.ac.ebi.quickgo.rest.search.results.PageInfo pageInfo = new PageInfo(1, 1, 5);
+        uk.ac.ebi.quickgo.rest.search.results.PageInfo pageInfo = new PageInfo.Builder()
```
borkabu: It does not require full package spec.

rest-common/src/test/java/uk/ac/ebi/quickgo/rest/search/solr/SolrQueryConverterTest.java
```diff
@@ -218,6 +200,120 @@ public void convertQueryRequestWithNoProjectedFieldWillProjectNoFields() {
...
+        assertThat(sortClause.getItem(), is(sortField));
+        switch (sortClause.getOrder()) {
+            case desc:
+                assertThat(sortOrder, is(SortCriterion.SortOrder.DESC));
+                break;
+            case asc:
+                assertThat(sortOrder, is(SortCriterion.SortOrder.ASC));
+                break;
+            default:
+                throw new IllegalStateException("Could not verify sort criterion");
```
borkabu: I have a feeling that this exception is not tested anywhere.

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/query/CursorPage.java
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @param cursor the cursor from where to fetch the next result set
+     * @param pageSize the maximum number of results to include in the page of results
+     * @return the new instance
+     */
+    public static CursorPage createCursorPage(String cursor, int pageSize) {
+        return new CursorPage(cursor, pageSize);
+    }
+
+    public String getCursor() {
```
borkabu: Ignore, that was test.

rest-common/src/test/java/uk/ac/ebi/quickgo/rest/search/query/QueryRequestTest.java
borkabu: Probably ignore for now as I do not want to shadow Sonar.

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/solr/SolrQueryConverter.java
```diff
@@ -82,10 +83,54 @@ public SolrQueryConverter(String requestHandler, QueryVisitor<String> solrQueryS
...
-    private int calculateRowsFromPage(int page, int numRows) {
-        return (page - 1) * numRows;
+    private SolrQuery.ORDER sortOrderOf(SortCriterion.SortOrder sortOrder) {
+        switch (sortOrder) {
+            case ASC:
+                return SolrQuery.ORDER.asc;
+            case DESC:
+                return SolrQuery.ORDER.desc;
+            default:
+                throw new IllegalStateException("Unknown sort order specified");
```
borkabu: Not sure if that default behaviour is what desired. I would rather default to ASC. As it is now, there is no test for default case.

rest-common/src/test/java/uk/ac/ebi/quickgo/rest/search/query/QueryRequestTest.java
eddturner: no you're right thanks! I think these `throws` were not cleaned up after the tests had changed over time

rest-common/src/test/java/uk/ac/ebi/quickgo/rest/search/results/QueryResultTest.java
eddturner: ta

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/solr/SolrQueryConverter.java
eddturner: yes, annoying one that. The idea is that if in the future `SortOrder` was extended to include some other values, then this `default` branch would be reached, and the caller (our code) will have their application fail, with a clear message, at development time -- rather than getting a valid value, and probably logging a warning. What's also annoying is that it's not possible to test, unless I redefine `SortOrder` ...

Is it okay with you if I leave the exception in place, so that it is then the caller's responsibility to deal with the issue (e.g., add another switch branch) when they're doing development.

PR #134, opened 2017-01-25
Goa 2578 annotation service stream downloaded search results

eddturner:
Uses the Solr cursorMark approach iteratively to stream a succession of results via a service method. This will ultimately be used by a controller, but this PR focuses on just getting the service method doing the right thing.


rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/SearchDispatcher.java
```diff
@@ -111,7 +120,116 @@ private SearchDispatcher() { }
...
+                MutableValue<Integer> fetchedCount = new MutableValue<>(0);
+
+                int pageSize = firstQueryRequest.getPage().getPageSize();
+                int requiredIterations = getRequiredNumberOfPagesToFetch(
+                        pageSize,
+                        totalHits,
+                        limit);
+
+                resultStream = Stream.iterate(firstQueryResult, qr -> {
+                    String nextCursor = qr.getPageInfo().getNextCursor();
```
borkabu: That lambda is huge, lots of going on. Do not know if it is worth an effort to make it neater? For example, Have subclass StreamedQueryResult and have something clever in there... But do not dwell on that as this solution works.

rest-common/src/main/java/uk/ac/ebi/quickgo/rest/search/SearchDispatcher.java
```diff
@@ -1,51 +1,61 @@
...
+ * <p>
  * Validity check methods are also provided for common search parameters.
  * Note that this class is stateless.
- *
+ * <p>
  * Created 27/01/16
+ *
  * @author Edd
  */
 public final class SearchDispatcher {
```
borkabu: wasnt sure about everything here being static, but it seems that model is stateless, so I guess it is fine

PR #141, opened 2017-03-02
Goa 2675 search by goid needs defaults for go usage and go relationships

TonyWardell:


annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -586,45 +586,9 @@ public void idValidationTestWorksCorrectlyForGeneProductIDWithFeature() throws E
     }
 
     //---------- Gene Ontology Id
-
-    @Test
```
eddturner: Where are the tests for exact matching now and how would exact matching take place?

PR #142, opened 2017-03-02
Goa 2457 configurable co term test

eddturner:
It was noticed that the tests for annotation indexing (which includes testing co-terms), and co-term writing, wasn't maintaining control over where the file was being written to. Also, multiple test classes with different concerns were writing to the same files.

This PR puts the co-term specific properties into a separate class, which is populated through @ConfigurationProperties, and an instance of which is provided as a @Bean. The tests then are able to override this bean directly, and can write to a temporary location (BasicTemporaryFolder).

NB. This move to @ConfigurationProperties seemed relevant since we have an ongoing effort to use this style of property configuration over @Value, as recommended by the Spring community.

NB2. The use of @ConfigurationProperties means we'll need to change 1 property in our application*.yml files regarding coterm chunk size: `indexing.coterms.chunk.size` is now `indexing.coterms.chunkSize` (due to spring's naming conventions).

### DIFF
--- a/indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/coterms/CoTermsConfig.java
+++ b/indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/coterms/CoTermsConfig.java
@@ -28,7 +28,8 @@ import org.springframework.batch.item.file.transform.DelimitedLineAggregator;
 import org.springframework.batch.item.file.transform.LineAggregator;
 import org.springframework.batch.item.file.transform.PassThroughLineAggregator;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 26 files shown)


indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/coterms/CoTermsConfig.java
```diff
@@ -61,55 +65,64 @@
...
-    public Step coTermManualSummarizationStep() {
-        Preconditions.checkState(!allCoTermsPath.equals(manualCoTermsPath), "The output path for manual and all " +
-                "coterms files should not be the same, but they were both %s", manualCoTermsPath);
+    public Step coTermManualSummarizationStep(CoTermsConfigProperties coTermsConfigProperties) {
+        Preconditions.checkState(
+                !coTermsConfigProperties.getAll().equals(coTermsConfigProperties.getManual()),
+                "The output path for " +
+                        "manual and all " +
+                        "coterms files should not be the same, but they were both %s",
+                coTermsConfigProperties.getManual());
```
TonyWardell: We could put this preconditions check in the get method in the CoTermsConfigProperties class - just feels like better encapsulation .

eddturner: done

PR #145, opened 2017-03-09
Goa 2714 provide annotation date from about endpoint

TonyWardell:


annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationController.java
```diff
@@ -237,6 +246,18 @@ public AnnotationController(SearchService<Annotation> annotationSearchService,
...
+    /**
+     * Get meta data information about the Annotation service
+     *
+     * @return response with metadata information.
+     */
+    @ApiOperation(value = "Get meta data information about the Annotation service",
+            notes = "Annotations creation date.")
+    @RequestMapping(value = "/about", method = RequestMethod.GET, produces = {MediaType.APPLICATION_JSON_VALUE})
+    public ResponseEntity<MetaData> provideMetaData() {
+        return new ResponseEntity<>(metaDataProvider.lookupMetaData().get(0), HttpStatus.OK);
```
eddturner: `lookupMetaData` returns a list, e.g., if we want to provide multiple metadata like, GO version/timestamp and ECO version/timestamp -- if i understand correctly.

Given the ability to have multiple metadatum instances, I think it would beneficial if the response returned something like:

{
  "GO" : { "version" : "http://....", "timestamp" : "#e.g., 1/2/17"},
  "ECO" : { "version" : "http://....", "timestamp" : "#e.g., 1/2/17"},
  <anotherKey> : <flexibleType>
}

(or anything that will allow easy lookup and retrieval of multiple, possibly different metadata).

If I have a clear suggestion regarding serialization of the model, to produce something like the above, i'll post it here too.

eddturner: here's a snippet showing one way to do the flexible key/value approach:

https://gist.github.com/eddturner/3b5ec8cebcb5ac71617577a563c261e5

PR #144, opened 2017-03-09
Goa 2722 annotations taxon search with taxon usage exact or descendants

eddturner:
Note that the default behaviour (exact/descendants) can be changed by changing the default value used in AnnotationRequest. Initially it was set to exact, for consistency with the default behaviour in master, for go/ecoUsage.

However, it is now set to descendants, in _preparation_ for alignment to the new default of go/ecoUsage.

PR #146, opened 2017-03-09
Goa 2719 read and index new taxon ancestry data

eddturner:


indexing/src/main/java/uk/ac/ebi/quickgo/index/common/datafile/GOADataFileParsingHelper.java
```diff
@@ -66,4 +67,19 @@
...
     }
+
+    /**
+     * Splits a {@link String} value on occurrences of a {@link String} delimiter, into a list of {@link Integer}s.
+     * @param value the value to split
+     * @param delimiter the delimiter on which splitting takes place
+     * @return a list of {@link Integer} values
+     */
+    public static List<Integer> splitValueToIntegerList(String value, String delimiter) {
+        assert delimiter != null : "Delimiter cannot be null";
```
TonyWardell: asserts can be turned on or off - shouldn't this be a Preconditions.checkArgument? Asserts are used to confirm what otherwise can be assumed.

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
```diff
@@ -117,15 +116,29 @@ private int extractTaxonId(String rawTaxonId) {
...
+            if (matcher.matches()) {
+                return createNullableIntegerListFromDelimitedValues(rawAncestry, COMMA);
+            }
+        }
+        return singletonList(DEFAULT_TAXON);
+    }
+
+    private List<Integer> createNullableIntegerListFromDelimitedValues(String value, String delimiter) {
+        return value == null ? null : splitValueToIntegerList(value, delimiter);
+    }
```
TonyWardell: Could this method, adn the other createNullableStringList be pushed into GOADataFileParsingHelper?

For example GeneProductDocumentConverter does the same thing but differently (cut-and-paste below), so it would be nice to not repeat the code.

    @SafeVarargs private final <T> List<T> convertToList(T... elements) {
        List<T> list = Arrays.stream(elements)
                .filter(element -> element != null)
                .collect(Collectors.toList());

        return list.size() == 0 ? null : list;

indexing/src/test/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverterTest.java
```diff
@@ -287,6 +287,58 @@ public void convertsGoAspectAnnotationProperties() throws Exception {
...
+    // annotation properties: taxon ancestry
+    @Test
+    public void convertsNullAnnotationPropertiesToDefaultTaxonAncestryList() throws Exception {
+        annotation.annotationProperties = null;
+
+        AnnotationDocument doc = converter.process(annotation);
+
+        assertThat(doc.taxonAncestry, contains(DEFAULT_TAXON));
+    }
+
```
TonyWardell: Remove empty lines please. And in the methods below.

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
```diff
@@ -117,15 +116,29 @@ private int extractTaxonId(String rawTaxonId) {
...
 
+    private List<Integer> constructTaxonAncestry(String rawAncestry) {
+        if (!Strings.isNullOrEmpty(rawAncestry)) {
+            Matcher matcher = RAW_TAXON_ANCESTRY_REGEX.matcher(rawAncestry);
+            if (matcher.matches()) {
+                return createNullableIntegerListFromDelimitedValues(rawAncestry, COMMA);
+            }
+        }
+        return singletonList(DEFAULT_TAXON);
+    }
```
TonyWardell: I asked Tony S. about specifying a default taxon on 0. He says all annotations will have a taxon ancestry, so if it doesn't something has gone seriously wrong with the creation of the import files, and we need to abandon indexing.

indexing/src/test/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverterTest.java
```diff
@@ -287,6 +287,58 @@ public void convertsGoAspectAnnotationProperties() throws Exception {
...
+    @Test
+    public void convertsNullTaxonAncestryAnnotationPropertiesToDefaultTaxonAncestryList() throws Exception {
+        String value = null;
+        annotation.annotationProperties = buildKeyValuesPair(TAXON_ANCESTRY, value);
+
+        AnnotationDocument doc = converter.process(annotation);
+
+        assertThat(doc.taxonAncestry, contains(DEFAULT_TAXON));
+    }
```
TonyWardell: Given all annotations should have ancestry - this test should check for an exception - IllegalArgumentException?

indexing/src/test/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverterTest.java
```diff
@@ -287,6 +287,58 @@ public void convertsGoAspectAnnotationProperties() throws Exception {
...
+
+    @Test
+    public void convertsInvalidTaxonAncestryAnnotationPropertiesToDefaultTaxonAncestryList() throws Exception {
+        String value = "1234d";
+        annotation.annotationProperties = buildKeyValuesPair(TAXON_ANCESTRY, value);
+
+        AnnotationDocument doc = converter.process(annotation);
+
+        assertThat(doc.taxonAncestry, contains(DEFAULT_TAXON));
+    }
```
TonyWardell: As above.

indexing/src/test/java/uk/ac/ebi/quickgo/index/common/datafile/GOADataFileParsingHelperTest.java
```diff
@@ -116,6 +115,11 @@ public void splittingOnNullDelimiterCausesException() {
         splitValue("some value", null);
     }
 
+    @Test(expected = AssertionError.class)
```
TonyWardell: As above, I think we need to change exception type.

indexing/src/main/java/uk/ac/ebi/quickgo/index/common/datafile/GOADataFileParsingHelper.java
eddturner: Yes, I agree -- i'd followed the same pattern as was in `splitValue`. I think this was used before we'd adopted the usage of `checkArgument`. Changed both to use `checkArgument` now.

indexing/src/test/java/uk/ac/ebi/quickgo/index/common/datafile/GOADataFileParsingHelperTest.java
eddturner: done

indexing/src/test/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverterTest.java
eddturner: it's to split up the given/when/then, for visual clarity when reading the method (commented on in PR on taxon usage)

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
eddturner: Hmm, we'd had this conversation a long time ago, and the converse was said.

I'll leave this in so that data to ensure valid data ends up in Solr -- *but* but add explicit validation of the taxon ancestors properties, so that if there's nothing there, then indexing aborts.

indexing/src/test/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverterTest.java
eddturner: This test is testing a superset of documents that can enter the index. The behaviour of this test shouldn't happen because the validation framework prevents it from happening. This test then just makes sure that if something changes, or we reuse code, that only valid and searchable data is going to enter the index.

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
eddturner: Started doing this, but noting this involves a number of subtle changes and most importantly asserting and guaranteeing edge case behaviour during the splitting functions (e.g., `convertToList` has the slightly different behaviour to what is used in annotations; filtering null elements). 

looking closer, this should be in a different jira issue -- whose focus is to refactor common value splitting functions shared between geneproduct and annotation indexing, and ensuring edge-cases are handled.

indexing/src/test/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverterTest.java
eddturner: extra line indicates that we're doing the testing part of the test (given/when/then, etc)

eddturner:
Note that the refactoring of the variable *ANCESTRY to *ANCESTORS will take place during merging of this branch with https://github.com/ebi-uniprot/QuickGOBE/pull/144 (which makes applies the refactoring)

indexing/src/main/java/uk/ac/ebi/quickgo/index/annotation/AnnotationDocumentConverter.java
TonyWardell: I think you still need this test in AnnotationValidatorTest

    @Test(expected = ValidationException.class)
    public void invalidAnnotationPropertyDueToTaxonLinageContainingZeros() {
        annotation.annotationProperties = "go_evidence=IEA|taxon_id=35758|db_subset=TrEMBL|db_object_symbol=moeA5" +
                "|db_object_type=protein|taxon_lineage=0,0";
        validator.validate(annotation);
    }

At the moment the regex is allowing 0 as an OK taxon id

PR #148, opened 2017-03-15
Copy versioned files to target directory.

TonyWardell:


deployments/src/main/distros/rest/bin/copy-data
```diff
@@ -0,0 +1,115 @@
+#! /bin/bash
```
eddturner: This file should be made executable in the git repo, so that when it is cloned + copied, it can be run. `chmod 755 copy-data`

deployments/src/main/distros/rest/bin/copy-data
```diff
@@ -0,0 +1,115 @@
...
+
+##=========================================================================================
+# Copies files used by the QuickGO Backend application to a location that is profile
+# specific. Also backups up the previous version of the files used.
+##=========================================================================================
+
+set -euo pipefail
+IFS=$'\n\t '
+
+#THIS_DIR="$(dirname $0)"
```
eddturner: remove this commented-out line and the one below

deployments/src/main/distros/rest/bin/copy-data
```diff
@@ -0,0 +1,115 @@
...
+
+source "../common/common"
+
+# ======= read the variables used by the control scripts ================================
+source "environment.variables" || {
+    echo "Please create a file called, environment.variables, containing the necessary environment variables."
+    exit 1
+}
+
+readonly PROFILE="$1"
```
eddturner: The script fails with a not very clear error if no profile is passed. (Because $1 is empty)

changing this line to something like
```bash
readonly PROFILE=${1:-\'\'}
```

should ensure a default value is assigned to PROFILE, which should sort it out, though it'd need to be tested to double check the behaviour. 

The idea is that if the user of the script forgets to provide a profile, then a clearer error message is provided to user that reads something like:

`Input profile: '', not recognized. Allowable values: uni_qgo`

deployments/src/main/distros/rest/bin/copy-data
TonyWardell: Would you mind doing this as I have Windows?

deployments/src/main/distros/rest/bin/copy-data
TonyWardell: Done.

deployments/src/main/distros/rest/bin/copy-data
TonyWardell: Done, works well.

deployments/src/main/distros/rest/bin/copy-data
```diff
@@ -0,0 +1,112 @@
+#! /bin/bash
```
eddturner: Looking at this again, and the purpose of "recording file data at the time of indexing", that this script is run from within the REST script bin.

This is an example workflow that seems to be needed:
* indexing for annotation needs to take place
* run `copy-data` from the indexing script bin for annotation
* when annotation indexing is complete, restart annotation REST app

This point became apparent after last week's occurrence where it was unclear that annotation indexing needed to run REST's copy-data script.

Would there be any side-effect you can see if this was moved from the REST scripts to the indexing scripts?

TonyWardell:
copy-data script and copy files moved from rest to solr.

PR #168, opened 2017-07-21
A new parameter, extension_rel_db is used to take search requests in …

TonyWardell:
…the format 'relationship(DB)'

### DIFF
--- a/annotation-common/src/main/java/uk/ac/ebi/quickgo/annotation/common/AnnotationFields.java
+++ b/annotation-common/src/main/java/uk/ac/ebi/quickgo/annotation/common/AnnotationFields.java
@@ -57,6 +57,7 @@ public class AnnotationFields {
         public static final String TAXON_ANCESTORS = storeAndGet(VALUES, "taxonAncestors_unsorted");
         public static final String WITH_FROM = storeAndGet(VALUES, "withFrom_search");
         public static final String EXTENSION = storeAndGet(VALUES, "extension_search");
+        public static final String EXTENSION_REL_DB = storeAndGet(VALUES, "extension_rel_db_search");
 
         public static boolean isSearchable(String field) {
             return VALUES.contains(field);
--- a/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
+++ b/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
@@ -4,6 +4,7 @@ import uk.ac.ebi.quickgo.annotation.validation.service.ReferenceValidator;
 import uk.ac.ebi.quickgo.annotation.validation.service.WithFromValidator;
 import uk.ac.ebi.quickgo.common.validator.GeneProductIDList;
 import uk.ac.ebi.quickgo.rest.ParameterException;
+import uk.ac.ebi.quickgo.rest.controller.request.AllowableFacets;
 import uk.ac.ebi.quickgo.rest.controller.request.ArrayPattern;
 import uk.ac.ebi.quickgo.rest.search.AggregateFunction;
 import uk.ac.ebi.quickgo.rest.search.request.FilterRequest;
@@ -94,7 +95,8 @@ public class AnnotationRequest {
             Searchable.REFERENCE,
             Searchable.TARGET_SET,
             Searchable.WITH_FROM,
-            Searchable.EXTENSION
+            Searchable.EXTENSION,
+            Searchable.EXTENSION_REL_DB
     };
 
     /**
@@ -252,6 +254,11 @@ public class AnnotationRequest {
                     "results_in_formation_of(UBERON:0003070),occurs_in(CL:0000032),occurs_in(CL:0000008)," +
                     "results_in_formation_of(UBERON:0001675)")
     private String extension;
+
+    @ApiModelProperty(value ="Allow a search for a combination of relation and database without specifiying " +
+            "an id.",
+            example = "occurs_in(CL),transports_or_maintains_localization_of(UniProtKB)")
+    private String extensionRelDb;
 
     @ApiModelProperty(
             value = "The number of annotations to download. Note, the page size parameter [limit] will be ignored " +
@@ -564,6 +571,25 @@ public class AnnotationRequest {
     }
 
     /**
+     * A list of extension relationship to db search values, separated by commas
+     * In the format extensionRelDb=occurs_in(PomBase),acts_on_population_of(CL) etc.
+     */
+    public void setExtensionRelDb(String... relDb) {
+        filterMap.put(Searchable.EXTENSION_REL_DB, relDb);
+    }
+
+    /**
+     * Return a list of annotation extensionRelDb values, separated by commas
+     *
+     * @return String array containing comma separated list of extensionRelDb values.
+     */
+    @ArrayPattern(regexp = "(([\\w_]+)\\((([^,]+))\\))", flags = CASE_INSENSITIVE, paramName =
+            INCLUDE_FIELD_PARAM)
+    public String[] getExtensionRelD() {
+        return filterMap.get(Searchable.EXTENSION_REL_DB);
+    }
+
+    /**
      * Include fields whose values derive from external resources
      * @param includeFields a vararg of fields to include
      */
@@ -755,8 +781,10 @@ public class AnnotationRequest {
                 ", geneProductSubset='" + geneProductSubset + '\'' +
                 ", goIdEvidence='" + goIdEvidence + '\'' +
                 ", extension='" + extension + '\'' +
+                ", extensionRelDb='" + extensionRelDb + '\'' +
                 ", downloadLimit=" + downloadLimit +
                 ", includeFields=" + Arrays.toString(includeFields) +
+                ", selectedFields=" + Arrays.toString(selectedFields) +
                 ", filterMap=" + filterMap +
                 '}';
     }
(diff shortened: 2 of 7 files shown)


annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -1901,6 +1905,35 @@ public void resultsReturnedInOrderWrittenToSolrRelevancyNotUsed() throws Excepti
                 .andExpect(fieldInRowHasValue("geneProductId", 0, expected));
     }
 
+    //----- Tests for Annotation Extension Using Relationship and Db only ---------------------//
+
+    @Test
+    public void filterByExtensionRelationshipAndDbIdentifierOnly() throws Exception {
+//        final String searchString = ("results_in_development_of"); //$2
```
eddturner: are these comments needed -- looks like you can remove commented out lines?

solr-cores/src/main/cores/annotation/conf/schema.xml
```diff
@@ -287,6 +295,21 @@
         </analyzer>
     </fieldType>
 
+    <!-- Process searches for annotation extensions in the format 'relationship(db)'  i.e. without an identifier -->
+    <fieldType name="extension_rel_db" class="solr.TextField" positionIncrementGap="100">
+        <analyzer type="index">
+            <tokenizer class="solr.KeywordTokenizerFactory"/>
+            <filter class="solr.PatternCaptureGroupFilterFactory" pattern="([^,]+)" preserve_original="false"/>
+            <filter class="solr.PatternReplaceFilterFactory" pattern="([\w_]+)\(([A-Z]+):([\d]+)\)"
```
eddturner: does `[\w_]+` cover all dbs?

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -564,6 +571,25 @@ public void setExtension(String... extension) {
...
+    public void setExtensionRelDb(String... relDb) {
+        filterMap.put(Searchable.EXTENSION_REL_DB, relDb);
+    }
+
+    /**
+     * Return a list of annotation extensionRelDb values, separated by commas
+     *
+     * @return String array containing comma separated list of extensionRelDb values.
+     */
+    @ArrayPattern(regexp = "(([\\w_]+)\\((([^,]+))\\))", flags = CASE_INSENSITIVE, paramName =
```
eddturner: There should be tests for setting/getting this value, and validation of input values, as with all other fields.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -564,6 +571,25 @@ public void setExtension(String... extension) {
...
+            INCLUDE_FIELD_PARAM)
+    public String[] getExtensionRelD() {
```
eddturner: getExtensionRelD? or DB?

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -1901,6 +1905,35 @@ public void resultsReturnedInOrderWrittenToSolrRelevancyNotUsed() throws Excepti
...
+
+                ResultActions response = mockMvc.perform(get(RESOURCE_URL + "/search")
+                                                         .param(EXTENSION_REL_DB_PARAM.getName(), searchString));
+
+        response.andDo(print())
+                .andExpect(status().isOk())
+                .andExpect(contentTypeToBeJson())
+                .andExpect(totalNumOfResults(NUMBER_OF_GENERIC_DOCS))
+                .andExpect(fieldsInAllResultsExist(NUMBER_OF_GENERIC_DOCS));
+    }
```
eddturner: I think it'd be good to add some similar tests checking for some stranger db names, if there are any -- just double checking handling of any special characters used in db names, if there are any.

Same comment for strange identifier portions of the db:id.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
```diff
@@ -1901,6 +1905,35 @@ public void resultsReturnedInOrderWrittenToSolrRelevancyNotUsed() throws Excepti
...
+
+    @Test
+    public void filterByInvalidExtensionRelationshipAndDbIdentifierReturnsBadRequest() throws Exception {
```
eddturner: Could have one more test where wrongly spelt extension name, or with a strange character, finds nothing.

solr-cores/src/main/cores/annotation/conf/schema.xml
eddturner: is `[A-Z]+` able to capture all dbs?

eddturner: is `[\d]+` able to capture all ids?

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
TonyWardell: Removed - yet mean't to delete those.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
TonyWardell: Underscores.

solr-cores/src/main/cores/annotation/conf/schema.xml
TonyWardell: No - see above.

annotation-common/src/main/java/uk/ac/ebi/quickgo/annotation/common/AnnotationFields.java
```diff
@@ -57,6 +57,7 @@
         public static final String TAXON_ANCESTORS = storeAndGet(VALUES, "taxonAncestors_unsorted");
         public static final String WITH_FROM = storeAndGet(VALUES, "withFrom_search");
         public static final String EXTENSION = storeAndGet(VALUES, "extension_search");
+        public static final String EXTENSION_REL_DB = storeAndGet(VALUES, "extension_rel_db_search");
```
eddturner: Ideally, it'd be much better to find a way of searching on such values "rel(db)", using the `EXTENSION` field. This is because it's then consistent with the way clients search on the other individual fields. 

The fact this additional field approach works is great; I think the biggest problem is that it is not transparent to the client: they no longer can just throw a search to a single annotation extension request parameter. They need to know what they're really requesting, before sending the correct request. In our case, requiring a bit of extra work from the front-end (which I think is a bit of a code-smell).

If some more time could be spent finding a consistent way of doing this, it'd be much better I think.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -94,7 +95,8 @@
             Searchable.REFERENCE,
             Searchable.TARGET_SET,
             Searchable.WITH_FROM,
-            Searchable.EXTENSION
+            Searchable.EXTENSION,
+            Searchable.EXTENSION_REL_DB
```
eddturner: Ideally, it'd be much better to find a way of searching on such values "rel(db)", using the EXTENSION field. This is because it's then consistent with the way clients search on the other individual fields.

The fact this additional field approach works is great; I think the problem is mostly that it is not transparent to the client: they no longer can just send a search value to a single annotation extension request parameter, `extension`. They need to know what they're really requesting, before sending the correct request (involving either `extension` or `extensionRelDb` (I hadn't fully appreciated this on Friday). In our case, requiring a bit of extra work from the front-end (which I think is a bit of a code-smell).

If some more time could be spent finding a consistent way of doing this, it'd be much better I think.

Whether there is an extra field in the index, that's fine; it'd just be clearer for a client if they could get this behaviour by interrogating a single request parameter.

solr-cores/src/main/cores/annotation/conf/schema.xml
TonyWardell: I've added lowercase letters, numbers, hypen and underscore to the potential db values - basically it could be anything but colon.

TonyWardell: Now changed to all non-whitespace, works with colons in the id (as exists).

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
TonyWardell: Done.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
TonyWardell: whoops - fixed

solr-cores/src/main/cores/annotation/conf/schema.xml
TonyWardell: It didn't work for my underscore example, new version does  and should work for all.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerIT.java
TonyWardell: Removed - yes should have done this earlier.

PR #171, opened 2017-08-10
Goa 2930 tsv download should show aspect

TonyWardell:


annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/converter/AnnotationToTSV.java
```diff
@@ -57,25 +57,38 @@ private void initialiseContentMappings() {
             }
             j.add(nullToEmptyString.apply(c.annotation.goId));
         });
+        selected2Content.put(GO_ASPECT_FIELD_NAME, (c, j) -> j.add(nullToEmptyString.apply(aspectAsString(c
+                                                                                                                  .annotation.goAspect))));
```
eddturner: reformat class

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/converter/Aspect.java
```diff
@@ -0,0 +1,35 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author Tony Wardell
+ * Date: 09/08/2017
+ * Time: 16:45
+ * Created with IntelliJ IDEA.
+ */
+public enum Aspect {
+
```
eddturner: remove extra newline

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/converter/Aspect.java
```diff
@@ -0,0 +1,35 @@
+package uk.ac.ebi.quickgo.annotation.download.converter;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author Tony Wardell
```
eddturner: class description saying where/how it's used (we've got a few of these (!) for aspect over the code base now I think). If you think it can be merged with one of the other Aspect classes, shoot, else fine.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/converter/Aspect.java
```diff
@@ -0,0 +1,35 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(Aspect.class);
+    private final String scientificName;
+    final String character;
+
+    Aspect(String scientificName, String character) {
+        this.scientificName = scientificName;
+        this.character = character;
+    }
+
+    public static Aspect fromScientificName(String scientificName) {
```
eddturner: tests for this? (since now it's public)

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/DownloadResponseVerifier.java
```diff
@@ -103,7 +103,7 @@ static ResultMatcher selectedFieldsExist(String[] expectedFields) {
     }
 
     static class TSVMandatoryFieldMatcher extends TypeSafeMatcher<String> {
```
eddturner: I still find reading this class confusing. It's not really a generic TSV matcher (TSV fields can be customised, so field count + mandatory indices don't really make sense without context), i.e., this matcher is tailored specifically to the tests. Any suggestions?

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/download/header/TsvHeaderCreatorTest.java
```diff
@@ -111,23 +111,24 @@ public void writeHeaderForFullListOfFieldsNotSlimmed() throws Exception {
...
-                        + TAXON_ID + "\t"
-                        + ASSIGNED_BY + "\t"
-                        + ANNOTATION_EXTENSION + "\t"
-                        + DATE + "\t"
-                        + TAXON_NAME + "\t"
-                        + GENE_PRODUCT_NAME + "\t"
-                        + GENE_PRODUCT_SYNONYMS + "\t"
-                        + GENE_PRODUCT_TYPE + "\n",
-                MediaType.TEXT_PLAIN);
+                                         + SYMBOL + "\t"
```
eddturner: Looking at the tests I couldn't see one for `TsvHeaderCreator.write`'s handling when the `emitter` produces an exception.

What happens in this case? Does the download stop?

eddturner: Also just noting that there are classes named `*TSV*` and `*Tsv*`. Pedantic, sorry I noticed too. Just letting you know.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -267,8 +267,8 @@
 
     @ApiModelProperty(
             value = "For TSV downloads only, fields to return.",
-            allowableValues = "geneProductId,symbol,qualifier,goId,goName,evidenceCode,goEvidence,reference,withFrom," +
-                    "taxonId,assignedBy,extensions,date,taxonName,synonym,name,type.")
+            allowableValues = "geneProductId,symbol,qualifier,goId,goaspect,goName,evidenceCode,goEvidence,reference," +
+                    "withFrom,taxonId,assignedBy,extensions,date,taxonName,synonym,name,type.")
```
eddturner: This hasn't been tested (same issue as last time; should be added to the corresponding validation request IT tests).

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/converter/AnnotationToTSV.java
TonyWardell: Done

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/converter/Aspect.java
TonyWardell: Managed to convert 3 versions of Aspect to 1.
Use Optional rather than exceptions for non-matching versions, which I think makes the calling code a lot cleaner.
Kept the Aspect version in sources, as this should disappear once we get rid of the graph?

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/converter/Aspect.java
TonyWardell: New Aspect class is tested.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/download/header/TsvHeaderCreatorTest.java
TonyWardell: Well spotted, amended.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
TonyWardell: Done.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/converter/Aspect.java
TonyWardell: Now replaced.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/DownloadResponseVerifier.java
TonyWardell: Remove and replace with a specific test seems to be what you require.

eddturner: I have no preference of the solution chosen, just noting what I see. The DownloadResponseVerifier could be used by others, and they might expect it to check the fields contained in a download response for TSV (as described by the name). Since the TSV operation is flexible, this class won't check the response correctly unless they mock in a non-flexible way. GAF/GPAD are a bit easier in that they're fixed formats ... Anyway, yes chat later, no worries.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/header/HeaderCreationException.java
```diff
@@ -1,21 +0,0 @@
-package uk.ac.ebi.quickgo.annotation.download.header;
-
-/**
- * Class whose use indicates a problem occurred whilst creating a file download header section.
- *
- * Created 31/01/17
- * @author Edd
- */
-class HeaderCreationException extends RuntimeException {
```
eddturner: Why delete this?

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/download/header/TSVHeaderCreatorTest.java
```diff
@@ -69,35 +77,29 @@ public void writeHeaderForSeveralSelectedFields() throws Exception {
...
+                + TSVHeaderCreator.GENE_PRODUCT_ID + "\n", MediaType.TEXT_PLAIN);
     }
 
-    @Test(expected = HeaderCreationException.class)
-    public void ioErrorWhenWritingHeaderCausesHeaderCreationException() throws Exception {
-        doThrow(HeaderCreationException.class).when(mockEmitter).send(anyObject(), any());
-        when(mockContent.getSelectedFields()).thenReturn(singletonList(TAXON_NAME_FIELD_NAME));
-
-        TsvHeaderCreator tsvHeaderCreator = new TsvHeaderCreator();
+    @Test(expected = IllegalStateException.class)
```
eddturner: Why is it now an IllegalState?

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/header/GeneTypeHeaderCreator.java
```diff
@@ -60,7 +60,7 @@ private void send(ResponseBodyEmitter emitter, String content) {
             emitter.send(PREFIX + content + "\n", MediaType.TEXT_PLAIN);
         } catch (IOException e) {
             String errorMessage = "Failed to send TSV download header";
-            HeaderCreationException headerCreationException = new HeaderCreationException(errorMessage, e);
+            IllegalStateException headerCreationException = new IllegalStateException(errorMessage, e);
```
eddturner: What happens when a download request takes place from the controller?

common/pom.xml
```diff
@@ -37,6 +37,10 @@
             <groupId>org.springframework</groupId>
             <artifactId>spring-beans</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.data</groupId>
+            <artifactId>spring-data-solr</artifactId>
```
eddturner: Is it poss to leave this dependency in the pom that requires it? Otherwise, it means all projects import these libraries, even though they don't use `spring-data-solr` (which we've tried to avoid in the past).

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/download/header/TSVHeaderCreatorTest.java
TonyWardell: I didn't think HeaderCreationException was achieving very much, was misnamed and could be dealt with in a more generic way.
It was created only when there was an IOException on send - so its not actually the creation but dispatch of the header that is failing. Also, with HeaderCreationException being a RuntimeException that is not dealt with explicitly by ResponseExceptionHandler, I presume the client is going to receive a 500 response anyway
The emitter send method creates a IllegalStateException for any other exceptions it encounters, so I thought to just use this.
On reflection, an IOException suggests the client isn't going to receive output anyway (which is why there is no 'failure to communicate with client' HTTP status code), perhaps we could create an internal unchecked communication failure exception to cover this situation, and at least have 
ResponseExceptionHandler pick it up and try and send something sensible to the client (whether it gets there or not).

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/header/GeneTypeHeaderCreator.java
TonyWardell: The exception will be caught by ResponseExceptionHandler

common/pom.xml
TonyWardell: Of course - done.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/header/HeaderCreationException.java
TonyWardell: See below.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/DownloadResponseVerifier.java
TonyWardell: DownloadResponseVerifierSelectedFields created.  
TSVMandatoryFieldMatcher renamed to TSVAllFieldsMatcher

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/download/header/TSVHeaderCreatorTest.java
TonyWardell: Exception no longer thrown, only logged.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/header/GeneTypeHeaderCreator.java
TonyWardell: Exception no longer thrown.

common/pom.xml
eddturner: I still see this in lots of poms. If it's in common, then it need not be in any module that includes (transitively) common.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequestValidationIT.java
```diff
@@ -842,7 +842,7 @@ public void invalidIncludedFieldProducesValidationError() {
     @Test
     public void validSelectedFieldsAreAllValidatedCorrectly() {
         String[] validSelectedFields =
-                ("geneProductId|symbol|qualifier|goId|goName|evidenceCode|goEvidence|reference" +
+                ("geneProductId|symbol|qualifier|goId|goaspect|goName|evidenceCode|goEvidence|reference" +
```
eddturner: `goaspect` or `goAspect`?

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/header/AbstractHeaderCreator.java
```diff
@@ -0,0 +1,30 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.web.servlet.mvc.method.annotation.ResponseBodyEmitter;
+
+/**
+ * @author Tony Wardell
+ * Date: 15/08/2017
+ * Time: 15:59
+ * Created with IntelliJ IDEA.
+ */
+public abstract class AbstractHeaderCreator implements HeaderCreator{
```
eddturner: no javadocs.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/header/AbstractHeaderCreator.java
```diff
@@ -0,0 +1,30 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(TSVHeaderCreator.class);
+
+    @Override public void write(ResponseBodyEmitter emitter, HeaderContent content) {
```
eddturner: tests?

common/pom.xml
TonyWardell: I must have really screwed up the search for `org.springframework.data` . All removed now I believe.

annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequestValidationIT.java
TonyWardell: As the validation is case insensitive, and the value gets turned into lower case anyway it doesn't actually matter, but changed to camel case anyway for consistency,

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/header/AbstractHeaderCreator.java
TonyWardell: Done.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/download/header/AbstractHeaderCreator.java
TonyWardell: Done.

PR #191, opened 2018-01-12
Make Swagger for /downloadStats visible.

TonyWardell:
The checking in Statistics service has changed to reject requests without filtering parameters.


annotation-rest/src/test/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationControllerStatisticsIT.java
```diff
@@ -106,11 +107,19 @@ public void setup() {
         repository.save(savedDocs);
     }
 
+    @Test
+    public void statsRequestFailsIfNoFilteringParametersAreIncluded() throws Exception {
+        ResultActions response = mockMvc.perform(get(STATS_ENDPOINT));
+
+        response.andExpect(status().isBadRequest());
```
eddturner: Could we check also the error message here, just to ensure the client knows what's wrong here?

PR #197, opened 2018-02-28
Goa 3151 new source for evidence filter presets

TonyWardell:
Use new file evidence_presets as the source,
The file is not gzipped so the loading  code had to be altered to cope.
Also added new preset item goEvidence as the existing field names were not a good fit.


quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/model/presets/PresetItem.java
```diff
@@ -125,7 +125,8 @@ public static Builder createWithName(String name) {
         DESCRIPTION("description"),
         NAME("name"),
         ID("id"),
-        URL("url");
+        URL("url"),
+        GO_EVIDENCE("goEvidence");
```
eddturner: introducing a specific value into this general class `PresetItem` goes against the grain and suggests all PresetItems can have a GO evidence. I would suggest either specialising PresetItem to handle this case, or leaving it as is. Possible?

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/RawEvidenceNamedPresetValidator.java
```diff
@@ -0,0 +1,26 @@
...
+ * Validates the {@link RawEvidenceNamedPreset} instances being read from flat-files and mapped via
+ * {@link StringToRawEvidenceNamedPresetMapper}.
+ *
+ * Created 05/03/18
+ * @author Tony Wardell
+ */
+public class RawEvidenceNamedPresetValidator implements ItemProcessor<RawEvidenceNamedPreset, RawEvidenceNamedPreset> {
+    @Override public RawEvidenceNamedPreset process(RawEvidenceNamedPreset rawNamedPreset) {
+        if (rawNamedPreset == null) {
+            throw new ValidationException("RawDBPreset cannot be null or empty");
```
eddturner: copy paste typo

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/ff/StringToRawNamedPresetMapper.java
```diff
@@ -17,14 +17,17 @@
  */
 public class StringToRawNamedPresetMapper implements FieldSetMapper<RawNamedPreset> {
     private final RawNamedPresetColumns rawNamedPresetColumns;
+    private final Supplier<RawNamedPreset> rawPresetSupplier;
 
-    public StringToRawNamedPresetMapper(RawNamedPresetColumns rawNamedPresetColumns) {
+    public StringToRawNamedPresetMapper(RawNamedPresetColumns rawNamedPresetColumns,
+            Supplier<RawNamedPreset> supplier) {
```
eddturner: Wondering why you provide a `Supplier` here? Shouldn't we also then make all instances passed as parameter a `Supplier` too?

eddturner: supplier variable name doesn't describe what it is, other than some kind of supplier. Can we rename this variable?

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/EvidencePresetsConfig.java
```diff
@@ -48,41 +46,53 @@ public Step evidenceStep(
...
     }
 
-    private ItemProcessor<RawNamedPreset, RawNamedPreset> rawPresetValidator() {
-        return new RawNamedPresetValidator();
+    private ItemProcessor<RawEvidenceNamedPreset, RawEvidenceNamedPreset> rawPresetValidator() {
+        return new RawEvidenceNamedPresetValidator();
     }
+
+    private RawEvidenceNamedPresetColumnsImpl getColumns() {
+        return (RawEvidenceNamedPresetColumnsImpl)
```
eddturner: Could we put this in `SourceColumnsFactory` for consistency, as with all other creations of `RawNamedPresetColumns`?

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/RawEvidenceNamedPresetColumnsBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+    public RawEvidenceNamedPresetColumnsBuilder(int namePosition) {
+        super(namePosition);
+        this.goEvidencePosition = UNINITIALIZED_POSITION;
+    }
+
+    public RawEvidenceNamedPresetColumnsImpl build() {
+        return new RawEvidenceNamedPresetColumnsImpl(this);
+    }
+
+    public RawEvidenceNamedPresetColumnsBuilder withGoEvidence(int _goEvidencePosition) {
```
eddturner: need the `_`? None of the other methods do this.

Sonar complains about this.

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/RawEvidenceNamedPresetColumnsBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+class RawEvidenceNamedPresetColumnsBuilder extends RawNamedPresetColumnsBuilder {
+    private int goEvidencePosition;
+
+    public RawEvidenceNamedPresetColumnsBuilder(int namePosition) {
+        super(namePosition);
+        this.goEvidencePosition = UNINITIALIZED_POSITION;
+    }
+
+    public RawEvidenceNamedPresetColumnsImpl build() {
```
eddturner: add `@Override`

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/RawEvidenceNamedPresetColumnsBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    static RawEvidenceNamedPresetColumnsBuilder createWithNamePosition(int namePosition) {
+        return new RawEvidenceNamedPresetColumnsBuilder(namePosition);
+    }
+
+    protected static class RawEvidenceNamedPresetColumnsImpl extends RawNamedPresetColumnsImpl {
+        private int goEvidencePosition;
+
+        protected RawEvidenceNamedPresetColumnsImpl(RawEvidenceNamedPresetColumnsBuilder builder) {
```
eddturner: needs only be package-private

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/RawEvidenceNamedPresetColumnsBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * A specialisation of {@link RawEvidenceNamedPresetColumnsBuilder} used to create
+ * {@link RawEvidenceNamedPresetColumnsImpl}
+ * instances, which capture information about the column indices relating to a {@link RawEvidenceNamedPreset}.
+ *
+ * Created 05/03/18
+ * @author Tony Wardell
+ */
+class RawEvidenceNamedPresetColumnsBuilder extends RawNamedPresetColumnsBuilder {
```
eddturner: no tests exist for this class

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/StringToRawEvidenceNamedPresetMapper.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * A specialisation of the {@link StringToRawNamedPresetMapper} for evidences, so the GO Evidence field can be
+ * correctly named.
+ * @author Tony Wardell
+ * Date: 05/03/2018
+ * Time: 10:13
+ * Created with IntelliJ IDEA.
+ */
+public class StringToRawEvidenceNamedPresetMapper implements FieldSetMapper<RawEvidenceNamedPreset> {
```
eddturner: no tests exist for this class

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/ff/SourceColumnsFactory.java
```diff
@@ -57,7 +51,6 @@ public static RawNamedPresetColumns createFor(Source source) {
 
     public enum Source {
         DB_COLUMNS,
-        ECO2GO_COLUMNS,
```
eddturner: I'd suggest just replacing this with a new one that you use, rather than in this one case performing the building of the `RawNamedPresetColumns` outside of this factory.

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/model/presets/PresetItem.java
TonyWardell: Subclassed PresetItem

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/EvidencePresetsConfig.java
TonyWardell: I left it out deliberately as TBH I don't see the point in SourceColumnsFactory. Each config knows what it needs to create, and SourceColumnsFactory is a class that knows about everything for some reason.
The SourceColumnsFactory.createFor method is static so its not as if the implementation can even be swapped at runtime.

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/RawEvidenceNamedPresetColumnsBuilder.java
TonyWardell: Removed - not sure what the hell I was thinking there.

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/RawEvidenceNamedPresetColumnsBuilder.java
TonyWardell: Done.

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/RawEvidenceNamedPresetColumnsBuilder.java
TonyWardell: Done.

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/ff/SourceColumnsFactory.java
TonyWardell: That would require the client to know they have to cast the value coming from SourceColumnsFactory, which I don't agree with.

eddturner: I see what you mean. Do you see what I mean? This change makes `SourceColumnsFactory` only used sometimes. So, creation of *PresetColumns instances are not in same class whose responsibility is knowing about these weird column indices. To me it looks a bit like a broken window that should be made consistent again somehow ...

Any thoughts?

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/EvidencePresetsConfig.java
```diff
@@ -48,41 +47,52 @@ public Step evidenceStep(
...
-        return new RawNamedPresetValidator();
+    private ItemProcessor<RawEvidenceNamedPreset, RawEvidenceNamedPreset> rawPresetValidator() {
+        return new RawEvidenceNamedPresetValidator();
     }
+
+    private RawEvidenceNamedPresetColumnsImpl getColumns() {
+        return RawEvidenceNamedPresetColumnsBuilder
+                        .createWithNamePosition(1)
+                        .withIdPosition(0)
+                .withGoEvidence(2)
```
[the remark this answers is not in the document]
TonyWardell: Done

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/StringToRawEvidenceNamedPresetMapper.java
TonyWardell: Done

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/ff/SourceColumnsFactory.java
TonyWardell: My solution is to use SourceColumnsFactory to generate the columns, but using a named static method just for evidences, to avoid any casting. 
What do you think?

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/evidence/EvidencePresetsConfig.java
eddturner: Ah, finally found the comment ... (it was collapsed and hidden).

The point of `SourceColumnsFactory` is to encapsulate the magic numbers (column numbers) used to read relevant info from files -- which are used in multiple configurations -- and these magic numbers are:
* confusing to newbies that pick this up. If all together, it encapsulates the column number problem, rather than spreading it out
* more importantly, the enum values used by `SourceColumnsFactory` are used by multiple configurations (e.g., assignedby/reference/withfrom, etc.) -- so rather than copy/pasting the same set of confusing magic numbers 3 times, just call the factory 3 times.

PR #198, opened 2018-03-01
Add regulates to the list of valid relationships to use for slim trav…

TonyWardell:
…ersal.


ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/model/OntologyRelationType.java
```diff
@@ -50,7 +50,7 @@
                     .map(OntologyRelationType::getByLongName)
                     .collect(Collectors.toList());
 
-    public static final String DEFAULT_SLIM_TRAVERSAL_TYPES_CSV = "is_a,part_of,occurs_in";
+    public static final String DEFAULT_SLIM_TRAVERSAL_TYPES_CSV = "is_a,part_of,occurs_in,regulates";
```
eddturner: In the Jira https://www.ebi.ac.uk/panda/jira/browse/GOA-3156, it says:

* Add regulates, positively_regulates and negatively regulates to the list of allowed relationships for slimming defined within ontology

But the relationships over which slimming takes place remains the same. See https://github.com/ebi-uniprot/QuickGOBE/blob/master/ontology-rest/src/main/java/uk/ac/ebi/quickgo/ontology/traversal/TermSlimmer.java.

Also, it'd be good to include some tests to demonstrate the required behaviours.

TonyWardell: Ah yes, after I wrote the jira I realized regulates covers all the regulates types.

I don't want to create unit tests for this - we are effectively changing data not functionality.

eddturner: We will need tests for this, to document the correct behaviour (made same comment in the past), so that we even better lock down the desired behaviour -- then if someone changes this in the future, then they're notified.

eddturner: also, the need for tests is also specified in the jira you wrote:
* Ensure that a request of regulates for relationship is interpreted as regulates, positively_regulates and negatively

TonyWardell: 2 tests added.

PR #199, opened 2018-03-02
Goa 3148 qualifier preset not should be uppercase

TonyWardell:


quickgo-client/src/test/java/uk/ac/ebi/quickgo/client/service/loader/presets/qualifier/QualifierPresetsConfigTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private CompositePresetImpl presetBuilder;
+
+    @Before
+    public void setUp() {
+        presetBuilder = new CompositePresetImpl();
+    }
+
+    @Test
+    public void lowercaseNotBecomesUpperCaseNot() throws Exception {
```
eddturner: using or not using camel case for lower/uppercase is good, just noticed that they're different though (lowercase + UpperCase). Really tiny, but noticed it.

quickgo-client/src/main/java/uk/ac/ebi/quickgo/client/service/loader/presets/qualifier/QualifierPresetsConfig.java
```diff
@@ -53,14 +53,18 @@ public Step qualifierStep(
...
-                        PresetItem.createWithName(rawItem.name)
+                        PresetItem.createWithName(uppercaseNots(rawItem.name))
                                 .withRelevancy(rawItem.relevancy)
                                 .build());
             });
         };
     }
+
+    private String uppercaseNots(String name) {
+        return name != null ? name.replace("not|", "NOT|") : null;
```
eddturner: Hmmm, actually, where is the 'not' coming from? Is it possible to change at the source?

TonyWardell: The annotations input file has the qualifier case correct e.g. NOT|enables
This value gets saved to 2 fields
qualifier is unchanged and unindexed - its the value that appears in the annotation list and casing is correct.
qualifier_unsorted is lowercased and indexed - its the value used to filter by qualifer

The preset service uses this url to grab the qualifiers list 
http://wp-p1m-c5:8090/solr/annotation/terms?q=*%3A*&terms.fl=qualifier_unsorted
so its retrieving values that have been lowercased

We could, apart from the current solution
- index qualifier and use this as the source of the preset data
- not lowercase qualifier_unsorted but allow case in-sensitive search 
- anything else?

eddturner: okay fine. i think go ahead with ur coded solution

PR #202, opened 2018-03-13
Goa 3147 change how presets work for with from and assigned by

TonyWardell:


solr-cores/src/main/cores/annotation/conf/schema.xml
```diff
@@ -264,6 +269,20 @@
         </analyzer>
     </fieldType>
 
+
+    <!-- Get database only from the form DB:XXXXXXX-->
+    <fieldType name="database_id" class="solr.TextField" positionIncrementGap="100">
+        <analyzer type="index">
+            <tokenizer class="solr.KeywordTokenizerFactory"/>
+            <charFilter class="solr.PatternReplaceCharFilterFactory"
+                        pattern="([^:]+).*" replacement="$1"/>
```
eddturner: so a db is always the first part before the ':' ? Can a db have a colon in its name?

I think it'd be good to have a few tests that test that the `replacement` is always the db -- otherwise we haven't got any tests that ensure the presets db info will definitely get hits.

quickgo-client/src/test/resources/application.yml
```diff
@@ -25,4 +25,12 @@ search:
           resourceFormat: "/solr/annotation/terms?q=*%3A*&wt=json&indent=true&terms.fl=qualifier",
           responseClass: "uk.ac.ebi.quickgo.client.service.loader.presets.RelevancyResponseType",
           responseConverter: "uk.ac.ebi.quickgo.client.service.loader.presets.RelevancyResponseConverter"
-        }
\ No newline at end of file
+        }
+      - signature: withFrom
```
eddturner: This could have a name indicating that it is to do with presets? Otherwise it looks like a typical filter that could be accessed from the controller layer. I'd suggest changing it, "withFromPresets" or something with a bit of description.

PR #207, opened 2018-05-03
Goa 3201 complex portal id search

TonyWardell:
The hyphen in ComplexPortal ids has stops the identifier being treated as a single entity within Solr - its is broken up by the StandardTokienizer.  I've used the ClassicTokenizer instead.

### DIFF
--- a/solr-cores/src/main/cores/geneproduct/conf/schema.xml
+++ b/solr-cores/src/main/cores/geneproduct/conf/schema.xml
@@ -150,18 +150,6 @@
         </analyzer>
     </fieldType>
 
-    <!--<fieldType name="text_ignorecase_classic" class="solr.TextField" positionIncrementGap="100">-->
-    <!--<analyzer type="index">-->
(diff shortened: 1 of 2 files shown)


solr-cores/src/main/cores/geneproduct/conf/schema.xml
```diff
@@ -141,16 +141,15 @@
 
     <fieldType name="text_ignorecase" class="solr.TextField" positionIncrementGap="100">
         <analyzer type="index">
-            <tokenizer class="solr.StandardTokenizerFactory"/>
+            <tokenizer class="solr.ClassicTokenizerFactory"/>
```
eddturner: If a complex id can be of the form: `[A-Za-z]+-[A-Za-z]+`, then classic tokenizer won't work. Is this possible?

TonyWardell: The complex portal regex is  now
^(?:ComplexPortal:)?(CPX-[0-9]+)$

PR #214, opened 2018-06-19
Goa 3232 new qualifier values

TonyWardell:


quickgo-client/src/test/java/uk/ac/ebi/quickgo/client/model/presets/impl/CompositePresetImplTest.java
```diff
@@ -202,6 +203,18 @@ public void presetsAreReturnedInRelevancyOrder() {
                     presets.stream().map(p -> p.getProperty(NAME)).collect(Collectors.toList());
             assertThat(presetsNames, Matchers.contains(name(1), name(2), name(3), name(4)));
         }
+
+        @Test
+        public void qualifiersAreReturnedInAlphabeticalOrder() {
```
eddturner: Could you add a few presets to the builder that include numbers, to make sure alphanumeric are tested? E.g., `A1`, `B1`, `A2` and `A10` -- to make sure it's doing the right thing in their presence.

PR #218, opened 2018-08-09
Goa 3252 download header link

rizwan-ishtiaq:
Three bug fixes
see below jira
GOA-3252
GOA-3263
GOA-3254


rizwan-ishtiaq:
Approved by edd

PR #236, opened 2019-07-22
Update spring

rizwan-ishtiaq:


annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequestBody.kt
```diff
@@ -0,0 +1,105 @@
...
+
+
+import uk.ac.ebi.quickgo.annotation.model.AnnotationRequest.DEFAULT_GO_USAGE
+import uk.ac.ebi.quickgo.annotation.model.AnnotationRequest.USAGE_RELATIONSHIP_PARAM
+import uk.ac.ebi.quickgo.rest.controller.request.ArrayPattern
+import uk.ac.ebi.quickgo.rest.controller.request.ArrayPattern.Flag.CASE_INSENSITIVE
+import javax.validation.Valid
+import javax.validation.constraints.Pattern
+
+data class AnnotationRequestBody @JvmOverloads constructor(@field:Valid var and: GoDescription? = null,
```
eddturner: Nice :).

I think it's worth probably having some javadoc/documentation here to explain that this class is a simple class with a builder, written using kotlin. and say why you decided to use kotlin.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/service/statistics/AnnotationStatisticsService.java
```diff
@@ -42,6 +42,7 @@
 
     private static final int FIRST_PAGE = 1;
     private static final int RESULTS_PER_PAGE = 0;
+    private static final String COLLECTION = "annotation";
```
eddturner: it might be a good idea to have a single constant used whenever you need to specify the collection name, so you don't need to have multiple constants with the same value.

you could have an enum / constants file, that is used by this class, and `AnnotationController`, etc.

annotation-common/src/main/java/uk/ac/ebi/quickgo/annotation/common/AnnotationDocument.java
```diff
@@ -1,17 +1,19 @@
...
 import java.util.List;
-import org.apache.solr.client.solrj.beans.Field;
 
 /**
  * Solr document class defining all fields within the annotation core.
  *
  * Created 14/04/16
  * @author Edd
  */
+@SolrDocument(collection = "annotation")
```
eddturner: `"annotation"` could be a constant used anywhere where you refer to this collection string.

annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/controller/AnnotationController.java
```diff
@@ -132,6 +132,7 @@
     private static final String GO_NAME = "goName";
     private static final String TAXON_NAME = "taxonName";
     private static final String EVIDENCE_NAME = "evidenceName";
+    private static final String COLLECTION = "annotation";
```
eddturner: `annotation` could be a common constant used elsewhere too

PR #248, opened 2020-09-15
provides download file type as a part of request param

rizwan-ishtiaq:
see for details
https://www.ebi.ac.uk/panda/jira/browse/GOA-3513

### DIFF
--- a/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
+++ b/annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
@@ -215,8 +215,8 @@ public class AnnotationRequest {
     private int page = DEFAULT_PAGE_NUMBER;
 
     @ApiModelProperty(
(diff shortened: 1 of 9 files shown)


annotation-rest/src/main/java/uk/ac/ebi/quickgo/annotation/model/AnnotationRequest.java
```diff
@@ -217,6 +214,13 @@
       allowableValues = "range[" + MIN_PAGE_NUMBER + "," + MAX_ANNOTATION_PAGE + "]", position = 25)
     private int page = DEFAULT_PAGE_NUMBER;
 
+    @ApiModelProperty(
+        value = "Only apply for download endpoint, pass this parameter when user can't provide accept header. when this" +
```
eddturner, line 218: -> Only applies for the download endpoint. Pass this parameter when user can't provide accept header. When this parameter is present, accept header is ignored.

eddturner (approved):
Just one very minor change to the api doc, which I've given a suggestion for.

Other than that, looks really good.

PR #249, opened 2021-02-03
Gaf2.2

rizwan-ishtiaq:
see for details
https://www.ebi.ac.uk/panda/jira/browse/GOA-3529
https://www.ebi.ac.uk/panda/jira/browse/GOA-3530


eddturner (approved):
Looks good!

